PR #3, opened 2019-11-16
Create proposal issue template

benhylau:


cjdelisle:
Maybe it should also delete this file? https://github.com/pkt-cash/ns-projects/blob/master/projects/template.md

benhylau:
@cjdelisle my mistake, please revert this commit.

PR #5, opened 2019-12-03
Bookkeeping: Added the relevant transactions to the pkt_insight proje…

cjdelisle:
…ct page

Not sure if we should require a PR for every change, but in any case here's a bookkeeping change to log the transactions.


cjdelisle:
I disabled the hard requirement for approvals of merges because it's clearly impossible to perform simple technical changes with that requirement. I agree though that we should seek approvals from the group before merging actual requests.

PR #7, opened 2019-12-16
Multi-Sig QT Wallet Project

artrepreneur:
A proposal for a full node QT standalone wallet application.

### DIFF
--- a/projects/2019_12_16_MultiSig_QT_Wallet.md
+++ b/projects/2019_12_16_MultiSig_QT_Wallet.md
@@ -1,0 +1,90 @@
+# Multi-Sig QT Wallet Project
+* Project Name: MultiSig_QT_Wallet
+* Contact Email: vishnu@gridfinity.com
+* Project participants:
+  * Vishnu Seesahai - Project Owner
+  * *Other participants will be added as needed*
+* Projected duration: 3 months
+* Projected effort: 1 person per month = 3 person-months
+* Requested PKT contribution: 40 million PKT
+* Project status: **PROPOSED**
+
+## Project summary
+
+The objectives of the Multi-Sig QT Wallet is to expand the PKT network adoption by users who don’t have the resources for using the current [text wallet](https://github.com/pkt-cash/pktwallet), increase the network's density of nodes, and improve security & control by the implementation of multi-signature.
+
+## 1. Project Relevance to the PKT Network
+> _"The [objective of The Network Steward](https://pkt-cash.github.io/www.pkt.cash/steward/) is to reduce barriers of entry to network / VPN providers in the pursuit of increased individual autonomy and privacy"_
+
+The creation of a new multi-sig QT PKT wallet can achieve the following benefits for the PKT Network:
+
+* Expand the adoption of the PKT Network to users who do not have the technical skills, time, or resources to operate the current [open-source files of a wallet](http://github.com/pkt-cash/pktwallet), and run a [full node](https://github.com/pkt-cash/pktd).
+* Higher security & control, backup feature, and joint-account capability, provided by the multi-sig-transaction accessibility.
+* Increase the density of the PKT Network nodes, to contribute to building the infrastructure of a functional decentralized network.
+
+In the list of examples of [Projects Supported by The Network Steward](http://github.com/pkt-cash/ns-projects/blob/master/readme.md), the first example is a wallet. As of the date of preparation of this proposal, the only available wallet is a [text-based wallet](http://github.com/pkt-cash/pktwallet).
+
+## 2. Project Deliverables
+The project will create new software for a standalone application, which should be downloaded and installed from an Installer. The installation will come bundled and ready to run a full-node, [PKTD](https://github.com/pkt-cash/pktd), as well as a [PKT Wallet](http://github.com/pkt-cash/pktwallet) instance.
+
+This deliverables are:
+* New software for a standalone application.
+* The software will be available under the [ISC](https://opensource.org/licenses/ISC) license.
+* Any artistic non-software assets will be available under the [CC BY 4.0](https://creativecommons.org/licenses/by/4.0/) license.
+* The maintainer of this software will be: Vishnu Seesahai
+* The software will be hosted on https://github.com/pkt-cash, if the network steward wishes it so.
+
+### 2.1 Development overview
+The development will be carried out by using the QT Software toolkit due to its history of reliability, portability, and low memory usage.
+
+The wallet will have a graphical user interface (GUI).
+Compatible operating systems: MacOS, Unix.
+The software will have future interoperability with Lightning Network.
+
+For the avoidance of doubt, the development will not be an SPV wallet, and it will not have features for colored coins. This software will not have a low-resource footprint because it will run a full node.
+
+### 2.2 Wallet target features
+1. Send and receive transactions.
+2. Send transactions to multiple addresses at once.
+3. Batch transactions for large UTXO input sets.
+4. Sign transactions.
+5. Verify transactions.
+6. Generate new regular addresses and new multi-sig addresses.
+7. Export private keys and import new private keys.
+8. Encrypt and decrypt messages.
+9. Password-protect the wallet.
+10. Save and restore the wallet backup from seed.
+
+### 2.3 Wireframe
+The following image is a wireframe of the application for illustration purposes. The design and location of the features may differ in the final version.
+<div style="max-width:1000px"> <media-tag src="https://files.cryptpad.fr/blob/49/491527e3bbf27b61ba591fa9d6b55df7fec38a010e2fbd38" data-crypto-key="cryptpad:k9IvGYHKMNLw1wxXPFQEIYuqTuu3UqDKKx9NsLUn32g="></media-tag> </div>
+ 
+### 2.4 Team qualifications
+Vishnu Seesahai is a mathematician and researcher specializing in engineering innovation across global markets, with a focus on stochastic processes, computing, and blockchains. He is also the founder of Healthmatica, a public benefit health tech. He is a graduate of Cornell with a BS in Computer Science and MS in Applied Mathematics from the Theory Center for Advanced Computing. He formerly served as SSWE at Ovid, the first health-tech company to offer full-text search of medical databases, and later CTO of Pixelpark AG, a subdivision of Bertelsmann. He is also the co-creator of the first DVAV (Digital Video Audio Village), purchased by Lionsgate Entertainment, which is currently in use on film/tv productions within their ecosystem.
+
+## 3. Success Criteria
+* At least one successful independent installation of the application.
+* At least 90% achievement of the described features.
+
+## 4. Payments
+All payments shall be made to `pHdMacckyFaTYqufm4GPDijeWzs9USt1Ju`
+
+## 5. Milestones
+The following are milestones for the progress of the project by which the network steward can evaluate the success of the project.
+
+### M0 - Kickoff
+At the kickoff of the project, the network steward will grant 20 million PKT to the applicant and the project will begin.
+
+### M1 - Minimum Viable Product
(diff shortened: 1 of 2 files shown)


projects/2019_12_16_MultiSig_QT_Wallet.md
```diff
@@ -0,0 +1,90 @@
...
+* Higher security & control, backup feature, and joint-account capability, provided by the multi-sig-transaction accessibility.
+* Increase the density of the PKT Network nodes, to contribute to building the infrastructure of a functional decentralized network.
+
+In the list of examples of [Projects Supported by The Network Steward](http://github.com/pkt-cash/ns-projects/blob/master/readme.md), the first example is a wallet. As of the date of preparation of this proposal, the only available wallet is a [text-based wallet](http://github.com/pkt-cash/pktwallet).
+
+## 2. Project Deliverables
+The project will create new software for a standalone application, which should be downloaded and installed from an Installer. The installation will come bundled and ready to run a full-node, [PKTD](https://github.com/pkt-cash/pktd), as well as a [PKT Wallet](http://github.com/pkt-cash/pktwallet) instance.
+
+This deliverables are:
+* New software for a standalone application.
```
benhylau: What operating systems will this be packaged for  / running on?

benhylau: Oops just saw it answered below.

projects/2019_12_16_MultiSig_QT_Wallet.md
```diff
@@ -0,0 +1,90 @@
...
+* The software will be available under the [ISC](https://opensource.org/licenses/ISC) license.
+* Any artistic non-software assets will be available under the [CC BY 4.0](https://creativecommons.org/licenses/by/4.0/) license.
+* The maintainer of this software will be: Vishnu Seesahai
+* The software will be hosted on https://github.com/pkt-cash, if the network steward wishes it so.
+
+### 2.1 Development overview
+The development will be carried out by using the QT Software toolkit due to its history of reliability, portability, and low memory usage.
+
+The wallet will have a graphical user interface (GUI).
+Compatible operating systems: MacOS, Unix.
```
benhylau: @cjdelisle I wonder what platform are people expected to run full nodes. Probably not MacOS based systems right?

If this is written in QT can it be targeted for different operating systems, with packages for Debian / Ubuntu, which I expect a lot of full nodes to use?

projects/2019_12_16_MultiSig_QT_Wallet.md
```diff
@@ -0,0 +1,90 @@
...
+5. Verify transactions.
+6. Generate new regular addresses and new multi-sig addresses.
+7. Export private keys and import new private keys.
+8. Encrypt and decrypt messages.
+9. Password-protect the wallet.
+10. Save and restore the wallet backup from seed.
+
+### 2.3 Wireframe
+The following image is a wireframe of the application for illustration purposes. The design and location of the features may differ in the final version.
+<div style="max-width:1000px"> <media-tag src="https://files.cryptpad.fr/blob/49/491527e3bbf27b61ba591fa9d6b55df7fec38a010e2fbd38" data-crypto-key="cryptpad:k9IvGYHKMNLw1wxXPFQEIYuqTuu3UqDKKx9NsLUn32g="></media-tag> </div>
```
benhylau: This doesn't seem to render.

projects/2019_12_16_MultiSig_QT_Wallet.md
cjdelisle: At this stage, we're not to the point where the chain is so huge that it's unreasonable to run a full node on one's laptop. Also because @artrepreneur is planning to use pktwallet codebase, whatever he develops will be forward compatible with the neutrino-backed light version of pktwallet whenever that gets up and running.

projects/2019_12_16_MultiSig_QT_Wallet.md
cjdelisle: That was a direct copy from CryptPad and CryptPad uses encrypted images which don't render outside. Anyway since I reviewed the pad I have the drawing here: https://cryptpad.fr/file/#/2/file/i7jC-Hb5h9xoLKr8fTnhZULF/

cjdelisle: @artrepreneur it would be a good idea to put the picture somewhere such as imgur.com and then add a link to that here since then it will appear for the future.

projects/2019_12_16_MultiSig_QT_Wallet.md
```diff
@@ -0,0 +1,90 @@
...
+
+### 2.2 Wallet target features
+1. Send and receive transactions.
+2. Send transactions to multiple addresses at once.
+3. Batch transactions for large UTXO input sets.
+4. Sign transactions.
+5. Verify transactions.
+6. Generate new regular addresses and new multi-sig addresses.
+7. Export private keys and import new private keys.
+8. Encrypt and decrypt messages.
```
benhylau: Can you clarify what are `messages` in this context? I also see a  ✉️message tab on the wireframe.

projects/2019_12_16_MultiSig_QT_Wallet.md
```diff
@@ -0,0 +1,90 @@
+# Multi-Sig QT Wallet Project
+* Project Name: MultiSig_QT_Wallet
+* Contact Email: vishnu@gridfinity.com
+* Project participants:
+  * Vishnu Seesahai - Project Owner
+  * *Other participants will be added as needed*
+* Projected duration: 3 months
+* Projected effort: 1 person per month = 3 person-months
+* Requested PKT contribution: 40 million PKT
```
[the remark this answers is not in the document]
benhylau: Noting that the proposal also commits to continued maintenance.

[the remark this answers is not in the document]
artrepreneur: In absence of price discovery, on an open exchange or via. actual utility, we have no way to price this asset at all. If we used a seigniorage estimation, it would be valued closer to 1/ten thousandth of a penny ($.0000012), which is it's min value, without premium. The calc is based on: Price / Coin = Hashes / Block X  Block / Coins X KW / Hash X Price / KW. So, this isn't a land grab by any means, and the commitment to maintenance is assured.

projects/2019_12_16_MultiSig_QT_Wallet.md
artrepreneur: This feature is similar to electrum wallet, where you can paste in strings, ie. messages, and encrypt / decrypt them with your key pair. Not super useful for this version of the wallet. But, the groundwork is in place to execute encrypted chats from your wallet, similar to [BCH](https://news.bitcoin.com/newly-launched-keyport-platform-enables-bch-powered-encrypted-messaging/)

projects/2019_12_16_MultiSig_QT_Wallet.md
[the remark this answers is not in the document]
benhylau: I was doing an estimate based on potential market size and max number of PKT on the chain, which of course does not factor in the risk associated with an early stage project, but it does give an idea of what a PKT is valuated at _if_ the network is successful.

Here's one last message that is not inline I hope you can comment on, in case you missed it https://github.com/pkt-cash/ns-projects/pull/7#pullrequestreview-333886785 thanks!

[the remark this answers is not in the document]
artrepreneur: Yes, I can certainly respond to this. First, I must point out that it seems to be standard protocol for coin-bearing blockchains to offer a QT wallet, and not having one would seem atypical in the space and likely frowned upon. 

Furthermore, the UX provided by a QT wallet's GUI would be a drastic improvement in experience for non-technical users. This allows us to on-board users who would outright reject the CLI experience wrt. key management and wallet functionality. 

Additionally, PKT Wallet's CLI fails to implement account functionality robustly. Since, PKT Wallet is derived from BTC Wallet, this was obviously incomplete in BTC Wallet as well. I surmise that this is probably related to bitcoin-CLIs deprecation of accounts. However, if you wanted granular control over the sendtoaddress RPC, you need accounts working as expected. Also, sendtoaddress tends to move all transactional change to segwit addresses. This can be difficult to deal with if one executes a lot of transactions, and it totally ruins the balances on the PKT Explorer rich list. All in all, this means that we desire granular control over which addresses we spend from, in other words, we have to execute a raw transaction to make this happen. Obviously, this approach is not expedient, or accessible, for most users. Thus, having this feature available, in a easy-to-use interface, is ideal for user retention. 

Another point to mention is that any installations of the QT wallet expands the number of full nodes on the network, which in turn, improves the consistency of the network. Lastly, the existence of a QT wallet has the potential to propel OTC and block trades through accessibility and simplification of the multi-sig process. Currently, we are using electrum to sign transactions and that wallet is neither native to the PKT Cash chain nor fully integrated with it.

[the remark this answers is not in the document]
benhylau: These are definitely desirable features, and the NS has experienced usability friction points while signing the last project. @cjdelisle had to craft a Tx to spend from specific (about to be burned) accounts and other NS key holders had to use a custom script to sign blocks:

>granular control over which addresses we spend from

>simplification of the multi-sig process

I also feel okay about the requested PKT amount given the commitment to continued maintenance. This project is definitely beneficial to the PKT project and I am approving this proposal ✅

I'd still like to clarify my earlier question, which is perhaps asking something beyond the scope of your proposed project:

>For example, I have excess bandwidth at home and a Mac laptop. Now there is a user-friendly wallet binary, so I install it... and what's next?

So I now have a QT wallet installed, and many others have done so as well, so the consistency of the network is more secure and clarity around value of PKT. How do you see the path from this point forward to the beginning of a bandwidth market?

projects/2019_12_16_MultiSig_QT_Wallet.md
backupbrain: @cjdelisle probably MAcOS will be the least likely full-node system, but QT does compile for MacOS

projects/2019_12_16_MultiSig_QT_Wallet.md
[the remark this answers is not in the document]
artrepreneur: Yes the goal is to utilize the portability of QT to have this wallet available for multi-platform usage including unix and windows. 

Regarding your second inquiry, it's hard to speculate about that at this time. But, keep in mind that this wallet will be able to run @cjdelisle SPV wallet as well. This means that the wallet supports lnd/lightning, which is a requirement for the bandwidth marketplace. On the roadmap for future versions of the wallet are colored coins, which are also needed for the bandwidth marketplace.  

Furthermore, having a lightweight version of this wallet (SPV), will set up nicely for porting to mobile. One very likely scenario is that the SPV wallet is integrated into Anode's VPN marketplace app. As the VPN marketplace is the first step in creating the bandwidth marketplace, the wallet will be instrumental in this effort.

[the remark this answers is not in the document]
benhylau: Okay so my understanding here of the expected outcome is:

1. More people run full nodes -> consistent network + more stable PKT pricing
2. Roadmap for wallet -> run on more OSes + SPV on mobile platforms
3. Adoption / integration by first VPN provider -> beginning of a bandwidth marketplace

Of course these are speculative and not part of the scope of this proposal. I am trying to make sure we are all clear on where this project is heading and where this particular project fits into the roadmap.

Thanks for your answers and please comment here if there is any misunderstanding in what I wrote.

projects/2019_12_16_MultiSig_QT_Wallet.md
```diff
@@ -0,0 +1,90 @@
...
+Compatible operating systems: MacOS, Unix.
+The software will have future interoperability with Lightning Network.
+
+For the avoidance of doubt, the development will not be an SPV wallet, and it will not have features for colored coins. This software will not have a low-resource footprint because it will run a full node.
+
+### 2.2 Wallet target features
+1. Send and receive transactions.
+2. Send transactions to multiple addresses at once.
+3. Batch transactions for large UTXO input sets.
+4. Sign transactions.
```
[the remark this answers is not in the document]
cjdelisle: I would imagine the answer is yes, given it's in the name of the project

PR #8, opened 2019-12-24
Update 2019_12_21_wallet_whitepaper_vote.md

Arceliar:
Just updating my response after some clarification of how earmarking would work.


cjdelisle:
You can drop `Caleb: Left as is absent reply after clarification.` since that was fixed

cjdelisle:
LGTM

PR #12, opened 2020-02-28
adding_2020_02_28_cjdns_wifi_based_interface_proposal

EliasMesh:


cjdelisle:
To the NS team, I'll remind you that as per [the procedure](https://github.com/pkt-cash/ns-projects/blob/master/acceptance_process.md) all discussion of the acceptability or non-acceptability of a project must be done *in public*.

If you wish to comment without having your name attached, you may comment in [this pad](https://cryptpad.fr/code/#/2/code/view/rWjM-erzQwQzX1Pg5SGL1JqcGiE6ZffVeBco4V3xk34/present/) (ask me for the edit link).

The applicant may make updates to the proposed project until the deadline, after that date we should mechanically accept the PR but leave the proposal in a PENDING state in order to freeze the proposal and prevent any additional changes being made.

cjdelisle:
@EliasMesh I copied the content of this proposal into a cryptpad in order to review it in it's final form, please make changes to correct the look of the markdown syntax, here is a pad: https://cryptpad.fr/code/#/2/code/edit/P4217vtm8WJ4CplAlipwgS7R/

Note: You can change this pull request until Midnight UTC, do so by changing [your branch](https://github.com/EliasMesh/ns-projects/blob/patch-2/Proposal.%2028.02.%20Cjdns%20wifi%20based%20interface)

cjdelisle:

cjdelisle:
The votes have been counted and unfortunately your project was not selected. Inside of the file [vote_2020_02_29.py](https://github.com/pkt-cash/ns-projects/blob/master/vote_2020_02_29.py) you will find in that file also some comments regarding your application which may help you to improve it and you are encouraged to submit again in the next round.

```
   Hazard control             3.6
   Scope and use of resources 2.57
   Short term impact          2.25
   Risk control               3.0
   Long term impact           2.3
```

cjdelisle:

* 36mn budget
* 1/4 frontload
* 12 PM
* 3 mn/PM
* There is a missing milestone 1, there is a kickoff and then a list of requirements for completion. I think the NS should assume that there is a milestone 1 heading on line 52
* **Caleb:** This applicant requested project ideas, so this application is based on a rough spec which I provided, the time estimation and costing was their own
* This project is targetting an item which was suggested by the NS
* **Neil Alexander:** This project certainly has interesting and valiant goals, and I have no doubt that we will be very interested in Wi-Fi as the network matures. However, I would take note that building real-world demos, particularly on iOS and Android mobile devices, will be high risk due to the limitations of these platforms.

## Arceliar
* 'short': 0.75,   # This would be useful in a common use case near the edge of the network, especially on devices that don't support ibss/adhoc or meshpoint modes ...
* 'long': 0.1,    # ... but my understanding is that being a wifi client really does mean you're a *client*, you don't get to directly interact with other nodes without going through the AP, and most (all?) hardware+driver combinations do not support associating with multiple APs at once (or being an AP and a client at the same time). It's still useful for clients that are too low powered to route traffic or meaningfully participate in the network without offloading the crypto. Since this is only really applicable to leaf nodes at the edge of the network, and wirelessly connected leaf nodes are probably end-user devices. In the long run, I think any end user device that's fast enough for a comfortable user experience is probably going to be fast enough to handle the cryptography without needing to offload it to the wireless hardware, which seems to be the main advantage to using AP/client modes this way.
* 'scope': 0.17,  # 0.5 M/wm divided by 36/12 M/wm
* 'risk': 0.9,    # This is kind of uncharted territory as far as I know, so while it's not obvious that this will definitely succeed, I would be fairly surprised if it doesn't work out.
* 'hazard': 0.5   # As in the above project, I don't know if there are conflicts of interest associated with any of the team, and there's nothing in the application to indicate one way or the other. So I'm starting with the same baseline score as in the projects with known conflicts.

## Caleb
Automatically connecting to encrypted wireless access points is a significant step in the process of rolling out
an internet sharing device and then later local mesh networks which will provide last mile connections to the
bandwidth market. In the immediate term, the focus is on VPN technology so this is not immediately necessary, but
it will become necessary soon as the next step in the plan is to move into internet sharing. The time effort of
12 person-months seems within the realm of reason for the scope of the project. The project is inherently somewhat
risky because it contains research (finding a secure protocol which works on iPhone and Android), however the
success criteria are well defined and the first milestone proves the feasibility of the project so the Network
Steward probably won't lose more than the up-front contribution of 1/4 if the research proves fruitless.

## Adonis
I don't believe this is a critical piece of infrastructure yet.
When PKT networks are used to facilitate peer-to-peer ISPs, this will be a wonderful way of enabling consumers to build their own ISP infrastructure.
Great team though.

PR #15, opened 2020-02-29
adding 2020_02_29_Price_Display proposal

johnsonjh:


cjdelisle:

cjdelisle:

```
   Hazard control             2.5
   Scope and use of resources 1.29375
   Short term impact          2.6
   Risk control               2.1
   Long term impact           1.5
```

cjdelisle:

* 8mn budget
* 1/4 frontload
* 1.5 PM
* 5.33 mn/PM - highest cost/PM
* **Arc** I believe the seigniorage example given in this proposal is (unintentionally) misleading and overly simplistic. The example formula using hash rate and cost of energy makes sense with a bitcoin-like mining algorithm designed to be computationally hard, but it doesn't account for pkt's bandwidth-hard (and/or memory-hard) algorithm, or that the requirements and payouts for block mining or ann mining can differ significantly. Note that if the cost estimation remains accurate without accounting for bandwidth, then either the approximation is in error, or the mining algorithm has failed to be sufficiently bandwidth-hard (which kind of defeats the purpose of pkt). So in addition to the assumption that the hardware is a gridfinity miner, I believe assumptions would also need to be made about the available anns in the mining pool and the network available to the mining hardware. Considering that it's only an approximation to begin with (and one that's only valid when the margin between value and production cost is small), it would be very difficult to convince me that the seigniorage is accurate to within more than an order of magnitude. While I do expect that the general trends would probably be accurate (e.g. if the seigniorage price goes up, the value probably also goes up by a roughly similar factor), I think trying to display a price in e.g. USD would be actively harmful to the network. If people see the ticker and take the number seriously, but the number turns out to be inaccurate, then in the best case scenario it under-values the currency and may harm adoption. In the worst case scenario, it over values the currency and turns the project into an accidental NS-funded panzi scheme. As such, I believe it would be highly inappropriate to accept this proposal until the NS or interested parties have had time to more thoroughly research the problem.

## Arceliar
* 'short': 0.5,     # this would be 1.0 if the prices were accurate and realtime. Since seigniorage is an estimation (and not one I'm confident adjusts quickly enough to be useful in a realtime ticker -- difficulty doesn't retarget that fast), I'm penalizing this by half. I'm not sure if that's too much or too little, it's mostly to indicate that it leaves a lot to be desired.
* 'long': 0.1,      # the estimation becomes much less useful the moment anyone opens an exchange and starts providing those prices in real(er) time. Unless you're actually mining in the environment the seigniorage estimation is based on, in which case it's difficult to understate how useful that information is, but I feel like the miners will figure that out on their own without NS funding.
* 'scope': 0.09375, # 0.5 M/wm baseline divided by 8/1.5 M/wm for this project
* 'risk': 0.0,      # I posted a long comment about this in the cryptpad. Long story short: by my understanding, the estimation would require a lot of assumptions which weren't stated in the proposal. It's not obvious to me that the approximation would be broadly applicable enough to be useful after those assumptions are taken into account.
* 'hazard': 0.0     # Relating to the above, if there's a NS-funded ticker that's giving inaccurate price information, then that could lead to a lot of problems (ranging from discouraging adoption if we under-value the currency, to accidentally starting a panzi scheme if we over-value it)

## Caleb
A price calculator might be a neat bonus item but it is not critical to the development of a bandwidth
market and in the long term, this role will be filled by exchanges. The time effort of 1.5 PM for a price
ticker is high, but not ridiculously high, however 5.3mn PKT per PM is the highest price charged by any
application. There are not a lot of success criteria placed on the milestones, so the NS does not have very
much leverage to reject milestone reports as not meeting the success criteria. The proposal contains the text
"may include banners linking to other resources related to the PKT network, such as Gridfinity or a wallet
service" which appears as though it might benefit certain parties more than others.

## Adonis
This is an essential feature for the the future tracking of PKT value over time.
I believe this will enable analysts, historians, and marketers to speak more transparently about the value of PKT
In a future with bandwidth markets, this is also an essential feature. People will need to be able to equate PKT to real-world value. This project should be very simple and therefore not so costly.

PR #16, opened 2020-02-29
adding 2020_02_29_PKT_Faucet proposal

johnsonjh:


cjdelisle:

cjdelisle:

```
   Hazard control             1.6
   Scope and use of resources 1.15
   Short term impact          2.95
   Risk control               2.95
   Long term impact           0.7
```

cjdelisle:

* 16mn budget - plus 200k/month
* 3/16th frontload
* 12 PM
* 1.33 mn/PM
* Based on feature-set, 12 person-months appears high
* **Arc** So to summarize: the applicant proposes to distribute up to 2.4 M pkt over 1 year. The fee to do so is an additional 16 M pkt. That looks like about 87% overhead. If we consider this proposal to be analogous to donating to a charity that's going to redistribute the funds, then this gets [the lowest rating possible](https://www.charitywatch.org/our-charity-rating-process#how-grades-are-calculated). Since the applicant and members of the project team are also known by some NS members (which *really* needs to be mentioned in the proposals themselves moving forward), awarding a project that uses funds so inefficiently would give a strong appearance of nepotism. It's honestly difficult for me to overstate how unacceptable this is. Let me put it this way: at work, if I saw someone awarding a contract based on a proposal that looked anything like this, I'd be obligated to report this to the office of internal audits so they can open up a formal fraud investigation with the state auditor's office. Let me be clear -- I'm not saying that the faucet idea itself is bad, but in its current form this is **not OK**, and frankly speaking, this would be inexcusable if not for the fact that this is the first round of proposals, I would just have to assume that the applicant is trying to defraud the NS and disregard any future applications (or quit if the NS makes it a habit of approving something so transparently hazardous -- that's not what I signed up for, and it really wouldn't be worth my time anymore).
* **Caleb** I like the idea of using the charity model to evaluate this proposal

## Arceliar
* 'short': 0.75,  # This is a nice project from a short-term perspective. It gives people a way to get a few pkt without mining, knowing someone they can buy/trade/be-gifted pkt from, or waiting for an exchange to open. Since cryptocurrency isn't a new concept anymore, I don't think it would be as useful as it was with e.g. bitcoin, but generally speaking for now I'd strongly prefer having a faucet over not.
* 'long': 0.0,    # Assuming we don't fund it forever, this eventually becomes completely useless, +- whatever long-term benefit comes from redistributing the code (which could be useful to other projects, but probably not to pkt)
* 'scope': 0.05,  # 0.5 M/wm baseline, divided by 16/12 M/wm, gives 0.375. However, only 2.4 / (16+2.4) ~= 13% of the budget is actually going to be distributed by the faucet, and the rest goes to developing the faucet. That seems like an insane overhead. I'm not sure how to really account for that, since I don't really know what an appropriate overhead would be. My best approximation is to just multiply by the 2.4/18.4 ~= 0.13 factor to account for the fact that we'd be paying so much to distribute so little, which seems like a very inefficient use of funds (ignoring the optics of the situation, for the moment, that goes under hazard).
* 'risk': 0.95,   # If funded, I wouldn't expect this project to have much risk of failure... it's been done before, so it's not exactly breaking new ground.
* 'hazard': 0.0   # Tied to the scope, I consider it unacceptable to pay anyone so much to distribute so little, it just looks really irresponsible at best. It looks transparently fraudulent when we consider that the applicant and team members are known by some NS members.

## Caleb
A way for people to acquire a small amount of PKT is indeed an interesting idea, however PacketCrypt was
designed in such a way as to allow casual mining of small amounts of PKT using announcement mining so this
should not be necessary. There is also evidence that crypto faucets can be run as successful businesses so
this project does not bypass a tragedy-of-the-commons problem which blocks the natural emergence of a piece
of key infrastructure. Development of an app of this complexity should take 1 to 3 weeks of developer time.
The 12 person-months quoted is unacceptable for reason of "significant damage to the perceived legitimacy of
the Network Steward" because accepting this project will appear as a handout to the applicant.

## Adonis
I believe the faucet should drip PKT from a testnet that resets every 24 hours, rather than from the actual PKT chain.
This is an essential feature for PKT developers and testers.
This should be a very simple project and therefore not so costly.

PR #18, opened 2020-02-29
adding 2020_02_29_Cloud_Computing proposal

johnsonjh:


cjdelisle:

cjdelisle:

```
   Hazard control             1.3
   Scope and use of resources 2.4
   Short term impact          2.2
   Risk control               1.45
   Long term impact           3.4
```

cjdelisle:

* 30mn budget
* 1/6th frontload
* 60 PM (declared 60-120 PM)
* 0.5 mn/PM (0.25 mn/PM if the time spent is 120 PM) - lowest cost/PM
* Notice the phrase: "The software will be available under applicable open source licenses when possible"
* **Arc** Unless the applicants have cracked the secret to high-speed fully holomorphic encryption, and found a way to transparently run existing code in such an environment (in which case, money is probably the least of their worries and we shouldn't be wasting funds on them), then I don't see this being up to par (never mind better) from a safety perspective. I'm going to go out on a limb here and guess that most users (businesses) will have at least some private information or proprietary business logic that they need to run on their servers. If they run their code on one of the major existing cloud computing platforms, then they can be reasonably sure that the provider won't steal their secrets -- since, if they did, it would be easy to bring a lawsuit against them. It's much harder to sue someone when they're just some random node in a distributed network and payed by a cryptocurrency, and that lack of obvious legal recourse means it could be a very hard sell for anyone with anything secret involved in their workflow. That said, anecdotally, I can think of a workflow where the most computationally expensive part (by a hilarious margin) should be fine in a network like this. So I think this is definitely still useful to have, just probably not to quite the extent that the applicants imply. I just wanted to mention that aloud in case anyone else hadn't thought about it and wants to take that into account when scoring the project.

## Arceliar
* 'short': 1.0,   # If executed successfully, this proposal gives people an incentive to use pkt, something to spend pkt on, and the kinds of things the users will do with it are likely to contribute to further growth. Parts of the code (related to payment processing etc) are potentially of value to other projects as well, if only as working examples.
* 'long': 1.0,    # This is in keeping with the general theme of separating service providers from infrastructure itself, and distributing ownership and operation of the infrastructure across the community at large.
* 'scope': 1.0,   # By the cost per work-month metric that I'm using now, for lack of better ideas, this is the most cost effective proposal. So it's the standard by which I'm judging the other projects for the time being.
* 'risk': 0.05,    # Unless the applicants have cracked the secret to high-speed fully holomorphic encryption, and found a way to transparently run existing code in such an environment (in which case, money is probably the least of their worries and we shouldn't be wasting funds on them), then I don't see this being up to par (never mind better) from a safety perspective. I'm going to go out on a limb here and guess that most users (businesses) will have at least some private information or proprietary business logic that they need to run on their servers. If they run their code on one of the major existing cloud computing platforms, then they can be reasonably sure that the provider won't steal their secrets -- since, if they did, it would be easy to bring a lawsuit against them. It's much harder to sue someone when they're just some random node in a distributed network and payed by a cryptocurrency, and that lack of obvious legal recourse means it could be a very hard sell for anyone with anything secret involved in their workflow -- I don't doubt that somebody can find a use case for this, but currently I have no reason to believe that such a use case would be of any particular benefit with regard to the goals of the pkt project and the NS.
* 'hazard': 0.0     # To be blunt, approving a project this large and risky, which was submitted by a team known to associate with NS members, would look far more nepotistic than what I can tolerate. There are definitely elements of this proposal that I like, so I would encourage the project team to consider reapplying with a smaller and safer subset of the project, convince us that they have a plan to ensure that compute nodes won't be able to steal their user's private data and/or code without being held accountable, or show us something to indicate that the demand for potentially-nefarious computing power is still high enough that the lack of accountability is unlikely to be an issue.

## Caleb
Though it is not in the immediate plan for the PKT ecosystem, this project may be considered useful to the
ecosystem because it creates another use case for PKT. It does not, however, directly align with the long
term goal of creating infrastructure necessary for the emergence of a bandwidth market.
There is an interesting research problem around automatic splitting of computation between multiple computers,
homogenous and heterogeneous. However this field remains in its infancy and is likely not worth the limited
resources of the Network Steward. The applicant has not expressed their intention to publish any papers and
the application does not alude to the development of any breakthrough technology, so it is likely that the
applicant intends to use off-the-shelf software such as Kubernetes, Docker and Tahoe-LAFS. Whatever the case
may be, since the applicant has made a commitment to performing fundamental research, the Network Steward must
assume that they intend to use existing technology. The development of an application with the features
described in the proposal should not be expected to require more than 4-6 person-months. With 60-120
person-months of effort and the declared scope, this project is unacceptable for the reason that it would
appear as a handout to the applicant. If the applicant wishes to re-apply, the Network Steward encourages them
to create a prototype, or apply for a project to create a prototype, such that they may have a working
demonstration to better clarify the project scope. It should be noted that the line "The software will be
available under applicable open source licenses when possible" creates uncertainty. Results of Network Steward
funded projects must be available to everyone in the ecosystem equally without unfairly benefitting any one
participant over any other. In cases when it is impossible to make certain components of a project open
source, applicants should be clear about exactly which components, why, and why the project is not unfairly
benefitting any one participant, i.e. the owner of the intellectual property.

## Adonis
I don't believe the time is right for this yet. 
In time, these services will be incredible, and they can be paid for in PKT and the services can be decentralized. 
I think the PKT technology and the consumer understanding of that technology must mature first.
I'm not sure this is attainable for the price stated.
This is an ambitious project that I think needs more time to develop.
There are a lot of moving parts including decentralized resource management, billing and account management, marketing and market making, and possibly PKT market making. 
I'm not convinced this project will not serve to benefit entirely a single private stakeholder

PR #19, opened 2020-03-01
RunePKTWallet

mehowtv:


cjdelisle:
Past the deadline, sorry

linfisher:
Hello Network Stewards,
We are very excited to have an opportunity to contribute solutions to help move PKT further in its development. 

I am writing at this time to ask for a full peer review by the Network Stewards relative to our failed submission. First, we are very respectful of the rules that govern the ecosystem. We are very aware and mindful of the need to adhere to given standards of practice that apply to all. 

That being said, we finalized our submission and were unsuccessful actually submitting it to GitHub. Primarily because I am NOT a coder. I am the business guy of our group and thought it would be a simple process. As simple as posting this message. Apparently not so... 

I did email it to Caleb and Jesse with a timestamp before midnight UTC. Caleb has since held the hard line on the rules.

We can fully appreciate that even a minute over should NOT be allowed in for the review. However, if anyone "else" gets their feathers ruffled, then we can point to technical difficulties and a submission on time that gave NO advantage to us over anyone else who submitted through GitHub. 

We are hereby respectfully requesting a special ruling on our submission to approve us for this round of project reviews. 

Thanks so much for your time and consideration. 
Lin Fisher

cjdelisle:
My answer is that in this case the rules cannot be bent at all. However the team @neilalexander @Arceliar @backupbrain @benhylau have the right to enter this discussion and overrule me.

linfisher:
Thank you, Caleb, for your review. I would only point out this rule seems to protect other proposals from any advantage gained. I would hope the decision of anyone else reviewing this would consider that no advantage was gained and there is legal evidence of the submission attempt and actual timestamp of the email. This will be my last comment as I am not seeking a back and forth, just a last point of clarity. Again thank you for the opportunity to propose solutions to your community.

benhylau:
Thanks @linfisher. My concern was to not set a precedence that future proposals could reference and say, "if that was allowed this should be okay too!" then it creates more unpredictability to our collective process, especially when this group hasn't gone through many cycles so taking the path of less unforeseen consequences seem prudent.

We also intend to have many cycles throughout the year, so the downside of just submitting in the next one hopefully isn't that much inconvenience.

Thanks for your intention to participate in the process.

PR #20, opened 2020-03-08
2020 02 29 tentitive results

cjdelisle:
I'd like everyone's final review of their votes before we merge this in, thanks.


cjdelisle:
I'm going to merge this in, it would be cool to have a little thumbs-up from @backupbrain and @Arceliar but I'm confident that everything is correct and everyone will anyway check again before paying.

PR #22, opened 2020-03-10
Added proposed changes to the process

cjdelisle:
I wanted to create a pull request to kick off discussion about changes to the process.

## Acceptance process changes
* Drop the complex bidding process in favor of simply allowing participants to change their person-month cost up until 48 hours after the initial deadline
* Added success criteria to the Risk criteria
* Added reporting of conflict of interest to the Hazard section
* Added a process for handling Milestone Reports

## Template changes
* Encourage applicants to specify *who* does the work
* Provide for declaration of pre-project work
* Added section for disclosure of conflicts of interest

## Readme
Fully re-wrote this as a sort of how-to for writing good projects

### DIFF
--- a/acceptance_process.md
+++ b/acceptance_process.md
@@ -3,39 +3,11 @@ The following is an approval process for PKT Network Steward funded projects. Fi
 
 * **Step 0**
   * Deadline for project proposals
-* **Step 1** (1 week after step 0)
-  * NS provides a ranking of proposals, winners are selected starting from the best and counting down the list, skipping any which cause the sum to go over the budget, and adding projects until the sum reaches at least 90% of the total budget or there are no more acceptable projects.
-    * **Note**: The reason to fill only 90% of the budget is to avoid incentivising a mass submission of micro-projects which bring little to no value. Remaining coins go to the next proposal window.
-  * Deadline for submission of "significantly similar" counter-proposals
-    * **Note**: Counter-proposals are submitted without knowledge of whether the project which they are countering will in fact be accepted
-* **Step 2**: (1 week after step 1)
-  * NS evaluates each counter-proposal according to evaluation criteria and compares it to the relevant original proposal, any counter-proposal which is not significantly similar and a significant improvement over the original proposal is eliminated and the original proposal wins.
-    **Note**: If there are multiple counter-proposals, they are ranked and all are immediately eliminated except for the best one
-  * Original applicants to provide amendments to original proposals to make them better than counter-proposals.
-    **Note**: Amendments to original applications are submitted without knowledge of whether the counter-proposal will be considered better
-    **Note 2**: If the counter-proposal is not accepted then the original project the one that wins, not the amendment.
-* **Step 3**: (1 week after step 2)
-  * NS evaluates amended original proposals and compares them to counter-proposals, any amended proposal which is not significantly similar and a significant improvement over the relevant counter-proposal is eliminated and the counter-proposal wins.
-  * Counter-proposers provide amendments to their counter-proposals, if a winner has not been selected at this point, the process continues at step 2.
-
-Once all competing proposals have been eliminated, the remaining proposal is deemed accepted and is funded. This happens even if other proposals remain in the amendment loop.
-
-If all involved applicants, counter-applicants and the NS team are in agreement, the amendment loop can happen faster than a 1 week period, but in case of non-response, 1 week is the deadline after which a participant is deemed to have failed to file an amendment to their proposal or counter-proposal.
-
-## Example of process
-1. Deadline 1: Alice submits a proposal
-* Deadline 2:
-    * Bob and Charlie, both thinking they can do the same project better/cheaper, submit counter-proposals
-    * The NS team provides their ranking and indicates that Alice's proposal has not been eliminated
-* Deadline 3:
-    * Alice submits an amended proposal which takes into account some of the improvements in the counter-proposals of Bob and Charlie
-    * The NS indicates that Bob's proposal is better than Alice's, Charlie's proposal being not better than Bob's is immediately eliminated
-* Deadline 4:
-    * Bob submits an amended counter-proposal
-    * The NS indicates that Alice's amended proposal is better than Bob's counter-proposal
-* Deadline 5:
-    * Alice submits a second amendment of her initial proposal
-    * The NS indicates that Alice's first amendment of her proposal is better than bob's amended counter-proposal, thus Bob's counter-proposal is eliminated and Alice's first amendment of her proposal wins
+* **Step 1** (48 hours after step 0 deadline)
+  * Deadline for anyone who wishes to change their person-month cost, allowing applicants to lower their
+  cost in response to another project.
+* **Step 2** (When reviewers complete their evaluation)
+  * Winners are selected and comments are made
 
 ## Evaluation criteria
 Each of the NS team members will score the proposals for each one of the evaluation criteria and then the [Reweighted Range Voting](https://www.rangevoting.org/RRV.html) algorithm (example counter provided here: [rrv_vote.py](https://github.com/pkt-cash/ns-projects/tree/master/rrv_vote.py)). This is done by scoring each proposal based on the criteria given below, with a combined score being calculated from.
@@ -68,6 +40,7 @@ Use of resources is the question of how "expensive" a project is and its ability
 Risk control is about the question of how likely a project is to fail at delivering the expected result. Projects which are inherently more risky will be evaluated more stringently on their proposed risk control. Specific questions useful to the evaluation of risk control include:
 * How evenly is the effort spaced over the milestones of the project ?
 * How evenly is the payment spaced over the milestones of the project ?
+* How well did the applicant specify success criteria for the project ?
 * To what extent are the risks defined and planned for in the project proposal ?
 * To what extent are the difficulties and potential blockers moved to the early parts of the project ?
 * What evidence is there that the project owner is capable of delivering ?
@@ -81,25 +54,16 @@ The objective of the Network Steward is to make funding available for projects w
   * The use of milestones, deliverables and evenly spaced payments conditioned on deliveries is encouraged.
 * To what extent does the project control the risk of arbitrage profit opportunity for the applicant ?
   * An example of arbitrage profit opportunity appears when there is a lack of market agreement on the cost of the deliverables which the project will produce. Logo design is a good example of a deliverable for which there lacks market agreement on the cost. While PepeiCo saw fit to spend one million dollars on their logo, the Nike logo was bought for a mere $35. In the case of such an expenditure, the NS team and overall community are unable to verify whether a project constitutes a million dollars worth of diligent effort or a $35 drawing sold at a markup.
+* Did the applicant declare and control for any potential conflicts of interest ?
 
-## Counter-proposal and Amendment Criteria
-Counter-proposals and amendments to existing proposals are compared using the 5 evaluation criteria but in addition, they must achieve the status of significantly similar and a significant improvement.
-
-### Significantly similar
-A counter-proposal or amendment should be considered to be significantly similar in the event that it targets the same objectives as the original proposal. Some questions which may be useful to the evaluation of significantly similar status include:
-* Do the two proposals have the same short and long term impacts ?
-* Assuming both projects were funded, would it be nonsensical for a single entity to use the results of both projects ?
-
-### A significant improvement
-The definition of a significant improvement is designed to prevent "stealing projects" by submission of proposals which are trivially less costly. While the exact definition of a significant improvement will be left to the discretion of the NS team, applicants can generally consider that:
-* A proposal of otherwise equal value but at more than 5% less cost is a significant improvement
-* A proposal which is not only better than the original, but in fact also better than another project which is better than the original is a significant improvement
-Counter-proposers should be aware that a copy/paste counter-proposal will generally not be considered to be equal to the original without a justification of domain knowledge equal to that which is justified by the original applicant through the text of the proposal.
-
-### Unacceptable projects
+## Unacceptable projects
 Some projects may be judged by the Network Steward to be unacceptable under any conditions. An intuitive example would be a proposal to build an assasination market which is blatantly illegal and contrary to the objectives of the project. These judgements will be made by each member of the NS team individually and projects which are not deemed acceptable by at least 3 of the 5 team members will not be funded, even if that means burning coins. Some questions which are useful in the evaluation of acceptability:
 
 * Does the project seem more likely to benefit the PKT ecosystem than to harm it ?
 * Does the project seem unlikely to cause significant damage to the perceived legitimacy of the Network Steward ?
 * Does the project seem likely to be a better use of resources than burning the coins ? (note that burning coins causes deflation which increases the value of all existing coins)
 
+## Milestone reports
+Milestone reports for ongoing projects must be submitted along with existing projects, the NS will have a
+simple majority vote on whether to accept the milestone and if it is accepted then it is paid out from
+the budget and the projects competing on the round are allocated the reduced round budget.
--- a/acceptance_process.md
+++ b/acceptance_process.md
@@ -65,5 +65,14 @@ Some projects may be judged by the Network Steward to be unacceptable under any 
 
 ## Milestone reports
 Milestone reports for ongoing projects must be submitted along with existing projects, the NS will have a
-simple majority vote on whether to accept the milestone and if it is accepted then it is paid out from
-the budget and the projects competing on the round are allocated the reduced round budget.
+simple majority vote on whether to accept the milestone and if it is accepted during a time when a project
+submission round is open, it will be paid out of the budget for the round, otherwise if there is no open
+project submission round, it will be paid from the Network Steward purse if possible, otherwise payment
+may be delayed.
+
+### Delayed milestones
+While it is not in the interest of the Network Steward to rush projects, it is important to guard against
+the possibility of applicants applying for projects and then not doing the work - planning to abandon the
+project unless there is a precipitous increase in the price at which they can sell PKT. The Network
+Steward may, at it's sole discression, stop payment on a project at any time, but the Network Steward
+*must* observe ongoing projects and verify that no project is being intentionally delayed.
--- a/acceptance_process.md
+++ b/acceptance_process.md
@@ -63,13 +63,6 @@ Some projects may be judged by the Network Steward to be unacceptable under any 
 * Does the project seem unlikely to cause significant damage to the perceived legitimacy of the Network Steward ?
 * Does the project seem likely to be a better use of resources than burning the coins ? (note that burning coins causes deflation which increases the value of all existing coins)
 
-## Milestone reports
-Milestone reports for ongoing projects must be submitted along with existing projects, the NS will have a
-simple majority vote on whether to accept the milestone and if it is accepted during a time when a project
-submission round is open, it will be paid out of the budget for the round, otherwise if there is no open
-project submission round, it will be paid from the Network Steward purse if possible, otherwise payment
-may be delayed.
-
 ### Delayed milestones
 While it is not in the interest of the Network Steward to rush projects, it is important to guard against
 the possibility of applicants applying for projects and then not doing the work - planning to abandon the
--- a/projects/template.md
+++ b/projects/template.md
@@ -1,6 +1,24 @@
-# Project Proposal Template
+# *Project Name*
 
-> **Instructions**: fill out the name of the project and an email of the person who will be responsible for it, this is the person who will be responsible for splitting up the awarded PKT between the participants. Other project participants may be named or may be pseudo-anonymous as long as there's some means of contact listed.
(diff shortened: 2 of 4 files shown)


cjdelisle:
Lets use this PR as a place to do public comments on the process and any changes we want to make.

acceptance_process.md
```diff
@@ -3,39 +3,11 @@ The following is an approval process for PKT Network Steward funded projects. Fi
...
-* Deadline 4:
-    * Bob submits an amended counter-proposal
-    * The NS indicates that Alice's amended proposal is better than Bob's counter-proposal
-* Deadline 5:
-    * Alice submits a second amendment of her initial proposal
-    * The NS indicates that Alice's first amendment of her proposal is better than bob's amended counter-proposal, thus Bob's counter-proposal is eliminated and Alice's first amendment of her proposal wins
+* **Step 1** (48 hours after step 0 deadline)
+  * Deadline for anyone who wishes to change their person-month cost, allowing applicants to lower their
+  cost in response to another project.
+* **Step 2** (When reviewers complete their evaluation)
```
benhylau: Would be useful to have some expectation of when

readme.md
```diff
@@ -3,50 +3,123 @@
...
+
+1. **Reduce your person-month cost**: If you can get the work done cheaper, or if you consider PKT
+to be worth more, then your project will stand out ahead of the others.
+2. **Reduce the "front-loading" of the project**: Asking for PKT before the first milestone is
+completed makes the project more risky and difficult to accept, the Network Steward will tend
+to prefer projects which ask for the least amount in the first period and push the majority of
+the payment back until after success-criteria are met.
+3. **Do pre-project work**: Spend time developing prototypes, mockups, even useful work on the
+project itself and report all of this using *negative milestones*. You write them the same way as
+ordinary milestones but you are writing success criteria for a success which you already achieved.
```
benhylau: While I understand there to be a certain amount of work in writing a proposal, I'd like to caution against placing too much emphasis on pre-project work because being able to spend so much time on a project without compensation is only affordable by very privileged people, and would place some strong candidates at a disadvantage or exclude them from participation altogether.

If someone has relevant background, show good understanding of the technical / social / economical challenges, and spent little time to put together a strong proposal, I wouldn't hold it against them compared to a project that has invested much more significant time into the project.

readme.md
```diff
@@ -3,50 +3,123 @@
...
+
+* Use a github
+[pull request](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/creating-a-pull-request)
+to submit the project to this repository
+* The file should be placed in the `/projects` folder
+* The name of the file should conform to
+[the naming convension](https://github.com/pkt-cash/ns-projects/tree/master/projects)
+
+### Early submission
+If you submit early then Network Steward team might provide feedback but others can also see
```
benhylau:
```suggestion
If you submit early then the Network Steward team might provide feedback but others can also see
```

readme.md
```diff
@@ -3,50 +3,123 @@
...
-  * Compatibility with Android and iPhone devices
\ No newline at end of file
+## Not a gravy train
+When you begin thinking about writing proposals, the first thing you need to understand is that
+writing Network Steward project proposals is not meant to be a profitable endevor. When you mine
+PKT, you're mining against all of the other miners, when you buy it in a private trade you're
+bidding against all of the other buyers, and when you apply for projects, you're competing against
+all of the other applicants.
+
+Therefore you should think about making a project proposal as a partnership which allows you to
```
benhylau:
```suggestion
Therefore you should think of making a project proposal as a partnership which allows you to
```

readme.md
cjdelisle: It's a good point that we need to be cautious of anything which would exclude participants. Some research grant-making organizations have very strict proposal processes with pages and pages of text to write for a proposal to be acceptable, this has the effect of severely limiting the applicants to a small pool and in this pool there is not a lot of competition which ends up being bad for the grant-maker. Avoiding this malady is the reason why I have intentionally kept a relatively simple proposal process, unfortunately applicants are still struggling with the proposal format - which is why I've embarked on writing this document.

Now if we were to imagine a person who was willing to work for 75% the cost of anyone else because they didn't need the money: as long as their project had no serious problems, our hands would essentially be tied in giving them the bid because it would otherwise be unfair advantage to whoever we gave it to instead.

If instead they did 25% of the work up-front, if we considered crediting the work which they had done already, we would be in the same situation, they're only asking for 75% as much as they could. In addition, people who don't have a lot of experience with writing proposals may be unable to clearly articulate the proposal but can make up for it by demonstrating a compelling proof of concept which shows what they intend to do (though we must be careful with proposals which make few promises).

Finally in the interest of the network as a whole, encouraging participants to develop prototypes - whether they will win the project or not - is an efficient way to leverage private investment to help build out the ecosystem.

So I am highly attentive to avoiding anything which might create a barrier of entry for otherwise good projects, I also want to note that any discrimination, even positive, constitutes some form of a gift to the applicant which IMO runs contrary to the "no unfair benefit" rule.

acceptance_process.md
```diff
@@ -68,6 +40,7 @@ Use of resources is the question of how "expensive" a project is and its ability
 Risk control is about the question of how likely a project is to fail at delivering the expected result. Projects which are inherently more risky will be evaluated more stringently on their proposed risk control. Specific questions useful to the evaluation of risk control include:
 * How evenly is the effort spaced over the milestones of the project ?
 * How evenly is the payment spaced over the milestones of the project ?
+* How well did the applicant specify success criteria for the project ?
```
neilalexander: Shouldn’t these ideally be templated or following guidance anyway?

acceptance_process.md
```diff
@@ -81,25 +54,16 @@ The objective of the Network Steward is to make funding available for projects w
   * The use of milestones, deliverables and evenly spaced payments conditioned on deliveries is encouraged.
 * To what extent does the project control the risk of arbitrage profit opportunity for the applicant ?
   * An example of arbitrage profit opportunity appears when there is a lack of market agreement on the cost of the deliverables which the project will produce. Logo design is a good example of a deliverable for which there lacks market agreement on the cost. While PepeiCo saw fit to spend one million dollars on their logo, the Nike logo was bought for a mere $35. In the case of such an expenditure, the NS team and overall community are unable to verify whether a project constitutes a million dollars worth of diligent effort or a $35 drawing sold at a markup.
+* Did the applicant declare and control for any potential conflicts of interest ?
```
neilalexander: Whilst no doubt important, I am wondering how easily this is verifiable?

acceptance_process.md
```diff
@@ -81,25 +54,16 @@ The objective of the Network Steward is to make funding available for projects w
...
+## Unacceptable projects
 Some projects may be judged by the Network Steward to be unacceptable under any conditions. An intuitive example would be a proposal to build an assasination market which is blatantly illegal and contrary to the objectives of the project. These judgements will be made by each member of the NS team individually and projects which are not deemed acceptable by at least 3 of the 5 team members will not be funded, even if that means burning coins. Some questions which are useful in the evaluation of acceptability:
 
 * Does the project seem more likely to benefit the PKT ecosystem than to harm it ?
 * Does the project seem unlikely to cause significant damage to the perceived legitimacy of the Network Steward ?
 * Does the project seem likely to be a better use of resources than burning the coins ? (note that burning coins causes deflation which increases the value of all existing coins)
 
+## Milestone reports
+Milestone reports for ongoing projects must be submitted along with existing projects, the NS will have a
+simple majority vote on whether to accept the milestone and if it is accepted then it is paid out from
```
neilalexander: If this is not already defined somewhere then a process should exist by which a milestone decision is reached, otherwise this reads very much like it is merely at the whim of the NS.

projects/template.md
```diff
@@ -67,10 +73,40 @@ After the project is accepted, the Network Steward will pay XXX PKT
...
+## Disclosure
+I hereby submit this application in good faith and I attest that I have made no effort, nor do I
+intend to make effort, influence the Network Steward to accept this or any other project I have
+submitted.
+
+*Please check one or more:*
+
+1. [ ] A company is receiving the funds
+  1. [ ] Company has financial relationships with one or more reviewers: *specify whom*
+  * [ ] Company has no financial relationships with any reviewers: *specify whom*
```
neilalexander: Odd to say that there is no relationship with any reviewers and then ask to specify?

projects/template.md
```diff
@@ -67,10 +73,40 @@ After the project is accepted, the Network Steward will pay XXX PKT
...
+
+## Disclosure
+I hereby submit this application in good faith and I attest that I have made no effort, nor do I
+intend to make effort, influence the Network Steward to accept this or any other project I have
+submitted.
+
+*Please check one or more:*
+
+1. [ ] A company is receiving the funds
+  1. [ ] Company has financial relationships with one or more reviewers: *specify whom*
```
neilalexander: “Financial or business connections” maybe?

projects/template.md
```diff
@@ -67,10 +73,40 @@ After the project is accepted, the Network Steward will pay XXX PKT
...
+  * [ ] Company has no financial relationships with any reviewers: *specify whom*
+* [ ] An individual is receiving the funds
+  1. [ ] Individual works for same company as one or more reviewers: *specify whom*
```
neilalexander: “Financial or business connections” perhaps here too.

projects/template.md
```diff
@@ -67,10 +73,40 @@ After the project is accepted, the Network Steward will pay XXX PKT
...
 ### Milestone 2
 
-...
\ No newline at end of file
+...
+
+## Disclosure
+I hereby submit this application in good faith and I attest that I have made no effort, nor do I
+intend to make effort, influence the Network Steward to accept this or any other project I have
+submitted.
```
Arceliar: We should maybe add a little more to this statement. It would be nice if there was also an attestation to the effect that, if the proposal is accepted, the applicant won't use their work on the project itself to do something detrimental to the pkt network. Some hypothetical bad-faith work on reasonable projects might be:
1. Writing documentation (e.g. whitepaper) which is misleading in a way that omits problems or over-states successes, in an effort to drive up the value of pkt.
2. Writing an application (wallet, exchange, etc.) with an intentional security vulnerability that allows the applicant to steal pkt (ok, this is probably illegal anyway, it's mostly to protect the NS if an applicant does something unexpectedly malicious and tries to play the "there's no rule that dogs can't play basketball" defense).
3. Writing an application (price ticker etc.) that has an intentional bug or design limitation which causes it to produce incorrect results, in a manner which would e.g. give incorrect information about pkt prices in an effort to manipulate the market. (See e.g. the 2015 underhanded c challenge "faking fissile material" for an example of how a "bug" can produce very specific and misleading incorrect results.)

readme.md
Arceliar: I think the main thing is that the applicant should include whatever they have that could convince us that they know what they're doing. If they've done any related work (prototyping some part of the project, for example), then including it makes sense. If they have something else they can show us, e.g. prior work that demonstrates competence with relevant skills, or documentation that indicates a solid understanding of the problem(s) to be solved by the project, then they should include that. The way I see it, there's an upper limit to how convinced we can be -- at some point, the chance of success goes to 100%, and it's becomes a matter of which project gives the most benefit for the least cost. So I don't particularly care if pre-project work is how they convince us of this, or if they do it some other way. That said, I don't think it makes sense to specifically mention pre-project work and omit mention of prior non-project-related work or other information that could demonstrate competence.

acceptance_process.md
```diff
@@ -3,39 +3,11 @@ The following is an approval process for PKT Network Steward funded projects. Fi
...
-* Deadline 3:
-    * Alice submits an amended proposal which takes into account some of the improvements in the counter-proposals of Bob and Charlie
-    * The NS indicates that Bob's proposal is better than Alice's, Charlie's proposal being not better than Bob's is immediately eliminated
-* Deadline 4:
-    * Bob submits an amended counter-proposal
-    * The NS indicates that Alice's amended proposal is better than Bob's counter-proposal
-* Deadline 5:
-    * Alice submits a second amendment of her initial proposal
-    * The NS indicates that Alice's first amendment of her proposal is better than bob's amended counter-proposal, thus Bob's counter-proposal is eliminated and Alice's first amendment of her proposal wins
+* **Step 1** (48 hours after step 0 deadline)
```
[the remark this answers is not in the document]
cjdelisle: Good point, this is not new though, people could submit similar proposals even with the counter-proposal process. The current band-aid solution is that I act as an air traffic controller for proposals because most of the applicants ask my advice on what they should work on. One reason why I'd like to bring in professional evaluators is because I would like to charge for my time consulting, but with me as an evaluator that is a more overt conflict than I am willing to accept - so currently I advise anyone who asks on a volunteer basis.

We should also consider that there is some "eggs in one basket" risk associated with accepting too many proposals from the same applicant, so I would be quite favorable to a mutual exclusion deduction which worked for both purposes, however I'm not sure we should block this round of changes to get it.

acceptance_process.md
cjdelisle: This is ultra-tricky, because if we set a deadline then what if we don't meet it ?

Getting voted out is something that can happen for any reason or no reason - normally if the coinholders think we're not taking our role seriously. But that is a really big deal, and other than that there just isn't any way for us to be punished.

Come to think of it, this is a great argument for us offloading all of the day-to-day work to others, because then most of the work will be done by someone with objectives, pay and success criteria.

acceptance_process.md
cjdelisle: It's hard to template success criteria because they're different for every project. That said, the rewrite of the readme document should be helpful in this regard. Anyway the reason I added it here is to make it clear that lack of good metrics for success should be considered a risk driver (as well as hazard).

projects/template.md
cjdelisle: I'm worried about "contract scope creep" here:

1. Someone either doesn't know what they're doing and they're just excited, or they do know and they will have no problem checking the box anyway (and will claim that it's the truth and is not misleading... then if they're completely discredited they will then fall back on "I didn't know")
2. Is a very clear case where a person who was going to do that would check the box anyway and it doesn't provide us any legal protection we didn't already have - either we had no idea they planned to put the bug and we are innocent or we did know and we're guilty, regardless of the box.
3. This is kind of the same question as 1, anyone who did this would claim they had no idea and the presence of a checkbox would not dissuade them.

So in all of these cases, the crook has plausible deniability and adding the checkboxes really only serves to increase effort and mental load on innocent people. We must not underestimate the anxiety which can be created, in honest people with no criminal intent, by things such as:  "[ ] I attest that I have not committed any violation of the Internal Revenue Code in the past 7 years"

What we could do is ask whether the project result will: "Present information which might lead to price speculation" and if the applicant says yes, they must add a paragraph explaining the controls which will be in place to prevent people being mislead by the information.

WDYT?

projects/template.md
cjdelisle: [Financial Conflict of Interest](https://www.washington.edu/research/compliance/financial-conflicts-of-interest-fcoi/) is already a defined term which includes remuneration (salary, contracting, etc) and stock or other equity holdings. Is there some other type of business that you were thinking of ?

projects/template.md
cjdelisle: Sounds good

readme.md
cjdelisle: I agree about "evidence of abilities", but if they did work on the specific topic I think that should also be credited because they are effectively asking for less contribution for their work. I'll add some notes about "evidence of abilities".

benhylau: I think the text for "evidence of abilities" and "prior non-project-related work" would help here, and we know that in our scoring, we should not hold it against proposals that are able to demonstrate that but not having too much of the project work already done, even though from a cost perspective it'd be nice if projects all come in 90% already done.

Prototyping is nice, but this shouldn't turn into a competition of "who has the most privilege to spend the most time working on something". I think this culture is a risk to the sustainability of the network.

cjdelisle: For me they're two different things:  "evidence of abilities" is related to the person or team executing the project and convincing us that they are low risk.  "pre-project effort" is basically effort that we should consider as having been done on the project even though it was done before the project was approved and it relates to the competitiveness of their bid.

benhylau: I think in the `risk` evaluation, I would like to treat these two things the same. You only need to do as much prototyping as to convince that you are not down a misguided path and what you propose if viable. I don't want to give the expectation that in order to win a project, you need to throw in more free labour than others.

In the `scope` evaluation, the project would cost less either because:

1. More of it is already done
2. You are very good at this and it takes you 1 hour to do what others will take hours to figure out

So in a way, we can look at the cost estimates objectively to rank the scope score.

cjdelisle: I can write a small bit of info about prototypes and prior work in order to make this clarification.

cjdelisle: done, please check

projects/template.md
cjdelisle: FYI I've added the "Present information which might lead to price speculation" checkbox in the most recent push.

projects/template.md
cjdelisle: Done, please check

acceptance_process.md
cjdelisle: So there are a couple of points here:

1. Delay: The issue of coins we "were going to pay" getting burned doesn't really work, because so far every payment we've made was at the moment when coins were just about to burn or were already burning. The problem I have with delay is that an applicant might propose a project and then wait to see if the price of PKT goes up and then deliver the project later if it does - so what they're getting is a [call option](https://www.investopedia.com/terms/c/call.asp) which is essentially free. There are two ways we could prevent this:
  * Kill projects which are delayed more than twice their projected timeline -- but by this rule the block explorer would have been killed, and I would like to believe it is a poster child of a well executed project. We don't really need projects to be delivered quickly and if we allow people lots of time then they might give us a lot more results in the end.
  * Re-value projects based on the current price of PKT -- but this requires a market to exist and it also raises the question: "what if the price goes down", in which case I think we would want to allow people to apply for an increase in the payout or abandon the project.
2. Unacceptable Milestones: If the applicant submits a milestone report which does not meet the success criteria laid out or otherwise does not represent what a reasonable person would consider to be a correctly delivered project, then we have a right to refuse it and insist on additional work or in a more extreme case, kill the project. There's really no way that this can be not "at the whim of the NS" because trying to codify what is a properly delivered project is opening up a [rules lawyering](https://en.wikipedia.org/wiki/Rules_lawyer) can of worms where unscrupulous applicants will attempt to push right up to the line without crossing it or even seek loopholes.

That said, I'm thinking we should probably accept and pay out project milestones even if there is no ongoing projects because it doesn't seem to take that long to validate a milestone report and for myself, I wouldn't mind getting paid out quickly, I'm sure others will feel the same.

acceptance_process.md
```diff
@@ -81,25 +54,25 @@ The objective of the Network Steward is to make funding available for projects w
...
-
-### Unacceptable projects
+## Unacceptable projects
 Some projects may be judged by the Network Steward to be unacceptable under any conditions. An intuitive example would be a proposal to build an assasination market which is blatantly illegal and contrary to the objectives of the project. These judgements will be made by each member of the NS team individually and projects which are not deemed acceptable by at least 3 of the 5 team members will not be funded, even if that means burning coins. Some questions which are useful in the evaluation of acceptability:
 
 * Does the project seem more likely to benefit the PKT ecosystem than to harm it ?
 * Does the project seem unlikely to cause significant damage to the perceived legitimacy of the Network Steward ?
 * Does the project seem likely to be a better use of resources than burning the coins ? (note that burning coins causes deflation which increases the value of all existing coins)
 
+## Milestone reports
```
neilalexander, line 66: This section feels a bit unclear and is a really long sentence. Any chance we can reword this?

projects/template.md
```diff
@@ -57,20 +82,65 @@
...
+> 2. How much of the budget should be paid out?
+>
+> If you have already done work on the project, create *negative* milestones and write success criteria for
+> these milestones in order to justify what has already been done.
 
 ### Milestone 0 (Kickoff)
 
-> **Instructions:** Explain how much PKT must be provided at the beginning of the project and any pre-existing deliverables which will be immediately delivered. **NOTE:** Projects which ask for more of their budget up-front will be regarded as more risky by the Network Steward, for details see the [acceptance_process](https://github.com/pkt-cash/ns-projects/blob/master/acceptance_process.md).
+> **Instructions:** Document anything which you did on the project before the project began, any prototypes,
+> mockups, planning or useful work which was done. This should corrispond to the **Pre-project effort**
```
neilalexander, line 95: spelling

projects/template.md
```diff
@@ -57,20 +82,65 @@
...
+
+## Disclosure
+I hereby submit this application in good faith and I attest that I have made no effort, nor do I
+intend to make effort, influence the Network Steward to accept this or any other project I have
+submitted.
+
+*Please check one or more:*
+
+1. Conflicts
+  1. [ ] A company is receiving the funds
```
neilalexander, line 123: The word “company” is vague here - suggest we maybe want to use “organisation” instead and specify that it may also cover private enterprise, charities, etc.

projects/template.md
cjdelisle: corrispond -> correspond

projects/template.md
cjdelisle: Changed, I'm fine with "Organization"

PR #23, opened 2020-03-19
Added milestone 1 report !!!!!

cjdelisle:
Been a long time coming but I think this is an excellent Milestone 1 for this project, I'm very proud of the results.

This was a collaboration between myself and @gorhgorh

Since I sit on the NS, I'll have to abstain from voting on this milestone though I clearly would not submit anything that I didn't think acceptable.

### DIFF
--- a/projects/2019_11_13_pkt_insight.md
+++ b/projects/2019_11_13_pkt_insight.md
@@ -143,7 +143,7 @@ action, payment or communication from the network steward at any time. For their
 applicant has no binding commitment or obligation at any time as a result of their participation
 in this project.
 
-## Project Status
+## Kickoff
 
 First payment of 10mn completed to address
 [p7Gdf7YhaxSkWm6u6yU452S6C9mJpuTfwu](https://pkt-insight.cjdns.fr/#/PKT/pkt/address/p7Gdf7YhaxSkWm6u6yU452S6C9mJpuTfwu)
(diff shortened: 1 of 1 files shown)


cjdelisle:
@neilalexander Time effort is about 30 days on my side and around 15 for gorhgorh (I need a timesheet), using 7 hour day, 134 hour/month formula which reflects the French schedule and holidays, this would come to 2.35 person-months. The reason why I didn't include that information in the report is because you have no way to verify that I'm telling the truth here, and the only tool you might have is to withhold funds because the project was cheaper than expected - but if you did that, it would be the last time any applicant ever gave you an honest data.

I foresee we will face situations where the results of the project do not reflect what was promised and look like something a student could do in a few days. In those cases we will need to decide whether we're going to withhold funds, but an applicant who has the audacity to deliver such a thing should have no problem to also fake the numbers, so they won't be any help to us there.

tl;dr I didn't include time because the information is untrustworthy and it creates poisonous incentives which reward crooked applicants and hinder honest ones

cjdelisle:
@Arceliar the bandwidth estimate is based on number of announcements in blocks mined in the period. Because announcements can be re-used, the formula for the estimate is as follows:

```
annoucement_count * log2(ann_min_diff) / log2(highest_ann_min_diff_today) / 60 * 1024 * 8 / 2
```

`annoucement_count * log2(ann_min_diff) / log2(highest_ann_min_diff_today)` is an estimate of the number of new announcements for the block period. It works because for the most part, announcement miners are mining all announcements of the same min work. That will likely change when multiple pools emerge but if they identify themselves in the coinbase field then we can continue computing this way.

`/ 60 * 1024 * 8` is per second, each announcement qualifies as 1024 bytes and each byte is 8 bits.

`/2` I don't remember why that's there, but because I have access to a block miner what I probably did was compare the results to `iftop` on my miner and I found that everything was coming out double so I halved it and called it good.

As of the time of this writing, the home page is saying 40mb and my block miner is averaging around 30.9mb which I find pretty good given the inherent fuzzyness of the "new announcements" estimation.

cjdelisle:
This has been partially paid out in transactions:

```
0fa922fdc76d6b9aae5e2512ebc97c49b7aeb41663a5b777f2a577cdd25a5e1e
e144a127075e0a9e32a0cd1d8b91d56d643cc99548b4433b2425584071a8434f
112cc4eee85092a6b16c9df41508c4cafd80892806aa50d22f3f1d1caa51e284
5b78d536f4ffc838ef2cae6cb55dc1ff9d8ff224c7542975a0009b709eca2b8c
134e14afe3dffeb499144b13005ea9c3456bcfeff3d3215304f75b713e6b4c76
d7dd72b3607bdc247ba598af2019db6a53ca8c347e5a94ecddcd1b2632847fdf
3f6eac52f326aab009568049c71dabeb87cf3f6ae559d91883521733a49d2973
0e8ba4f54d400b730bf10cee0150acde277049d70c50feef42c712ef2ff8aa80
1e9b93e54fa4eaf118571a60fc626394ea618015baa25af534cc488c9008001a
```

What remains is 0.5mn coins that were not paid because the transaction was not signed and broadcasted fast enough to avoid the burn limit and so that transaction had become unspendable before it could be sent.

The applicant (me) is ok with receiving this last payment later, when it is convenient for the NS to make it.

PR #25, opened 2020-05-21
Adding pkt-vpn-utility milestone 1 report

backupbrain:
Anode has an APK for download on Android. It connects to a single PKT-VPN server and surfs the internet from that exit node.


cjdelisle:
Merging because everyone who is elegable to vote agrees.

PR #29, opened 2020-06-26
Propose a new CFP round

cjdelisle:
What do you think ?  Time for another round ?


cjdelisle:
Merging this because 3 people agree, @Arceliar seems to not be available lately and I think @backupbrain is ok with it.

PR #30, opened 2020-06-29
Changed the aether mailing list to the pkt.chat server

cjdelisle:
...because aether mailing list is not being monitored.

Sorry for this, I didn't check over this part but we should not push people to a place where they won't get any feedback.


cjdelisle:
I'm going to merge this immediately assuming there will be consensus - because I don't want people to end up on a mailing list which is not being monitored. If there is any disagreement then I am happy to revert.

PR #31, opened 2020-07-01
Update 2019_12_14_pkt-whitepaper-website.md

josh-berger:
Here is the milestone 1 of the project. Please review and provide feedback.

### DIFF
--- a/projects/2019_12_14_pkt-whitepaper-website.md
+++ b/projects/2019_12_14_pkt-whitepaper-website.md
@@ -372,3 +372,6 @@ The applicant understands that the Network Steward is not a legal entity and no 
 [198](https://pkt-insight.cjdns.fr/#/PKT/pkt/tx/6912729413f16888cacacb006de0c3b3b70a3ad703bedfdc0bf5a478eaee2c57)
 [199](https://pkt-insight.cjdns.fr/#/PKT/pkt/tx/fcd0c40dccd650a271c9311b5ce302019455f8c02445faae1dee9e90b811a13f)
 [200](https://pkt-insight.cjdns.fr/#/PKT/pkt/tx/ec9ca663ea47b50aebaaa5b2d064959111ab09635cb26ecbad6bcb884cd34a8b)
+#Milestone 1 - Whitepaper v1
+First version of the white paper by Alex Lightman delivered for notes from the Network Steward. View the whitepeper here: https://docs.google.com/document/d/1tk81CIiKS6vElBx0IaKcJ8tyVx2Irot9IaMpcrtZVJw/edit?ts=5efccb42#heading=h.gjdgxs
+Please review and provide feedback.


cjdelisle:
After a discussion with the NS team, we reached the following conclusion:

The objective laid out in the project was to produce a document which "makes the complex comprehensible, the intangible graspable and the future urgent", we find that this submission is unacceptably sloppy and does none of these things. It is not the job of the Network Steward to educate the applicant in how a document should be written, so remark only on a few particular problems which we find representitive of the effort put into this document.

We observe that the markdown document is not properly formatted, the **Roadmap** section contains a broken and unreadable table, the links to the end-notes are incorrect, and the code sections are not formatted `as code`, but rather copy-pasted directly in. The Network Steward uses Github and Markdown for the management of documents, and while certain flexibilities may be afforded to otherwise excellent projects, we find that the effort in making the document in correct and beautiful Markdown is representitive of the effort in making the overall document.

The fact-checking of this document appears to be relatively non-existant, for example Benedict is referred to in the document as Benjamin, though a quick search of his online handle instantly reveals his correct name. Furthermore the **Benefits of the PKT blockchain** is wraught with claims which are highly misleading because, depending on interpretation, they are either falsehoods or are technically true but irrelevant to most people.

There is incorrect use of terminology, for example the phrase "Data is the new oil" is applied to internet access when in fact this term refers to the power which data affords to internet giants such as Google and Facebook. There is also unnecessary use of rare terminology such as "pumpamentals" - a term for which we have been unable to find a satisfactory definition but which sounds similar to a certain financial crime.

There is the appearence of many different source materials having been copy-pasted together in a haphazard way. Exhibit A and Exhibit C are essentially duplicates, text is copy-pasted from the readme document of https://github.com/cjdelisle/PacketCrypt rather than using a citation, and the phrase "Vodafone detecting a 50% rise in Internet use" is present twice just a few paragraphs apart.

Because of these and countless other similar issues, we have decided not to accept this milestone.

We advise the applicant that the failures described herein are only choice examples of the overall work and it is not the job of the Network Steward to hand-hold the applicant through the writing process, therefore if this document is re-submitted in substantially the same form, it will be summarily rejected.

There are two types of documents which the applicant may submit, one is an academically honest technical description of the project and the other is a more open-ended document of that which the project wishes to achieve. The easier of the two papers is the technical description because any computer science scholar with experience authoring peer-reviewed papers will be capable of giving a proper academic description of that which exists now. We foresee this paper being no more than 15 pages in length. An open-ended document of project objectives/vision is more difficult because it must describe technologies which do not exist yet while remaining within the realm of that which is possible - but the source materials which were copy/pasted into this paper can be used to inform such an effort. We foresee this paper being no more than 25 pages in length.

The Network Steward will accept either of these two works as fulfillment of this milestone.

PR #34, opened 2020-07-25
Submit tokenstrike proposal

cjdelisle:


cjdelisle:
I forgot to declare a conflict because Anode LLC has a financial relationship with Adonis as well as myself.

projects/2020_07_25_tokenstrike.md
```diff
@@ -0,0 +1,164 @@
...
+* Projected duration: 6 months
+* Projected effort: 7.2 person-months
+* Pre-project effort: 0.2 person-month
+* Requested PKT contribution: 40mn
+* PKT address to pay to: pkt1qju5ru8em5jz34r4g6htfvdp4cqjt4ey7ktmvv5
+
+## Project summary
+
+The objective of this project is to establish a protocol for the issuance and exchange of tokens. Tokens must be able to be issued and exchanged with minimal effort (gas fees needed for exhange of Ethereum tokens are unacceptable). A holder of tokens should be able to exchange them in a fully untrusted Lightning Network HTLC transaction. We relax the requirement that holders of tokens must be secure from attacks by the token issuer, and in it's place we add the requirement that any attack by the issuer of a token should be able to be proven to have taken place, such that an issuer who breaks protocol will be *discredited* and nodes will cease to interact with them until they are manually reset.
+
```
backupbrain, line 17: How do we know a discredited node has been manually reset?

cjdelisle: The discredited node isn't reset, everyone else is in order to not consider them discredited anymore

cjdelisle:
Since I wrote this proposal, it would be silly for me to say "congratulations me", but I'm going to write a comment here in order that those who look at this pull request later are able to see it in context.

This project was accepted in the round, the exact vote is present here: https://github.com/pkt-cash/ns-projects/blob/master/vote_2020_06_25.py but for convenience, the reviews are copied here.

## Reviewer 1
I believe this project will deliver fascitating improvements to PKT, including opening token swapping and lightning payments, which are directly in line with PKT's objectives.
* 'short': 0.1, I believe there is little short-term gain from implementing this project
* 'long': 0.9, I believe an advanced token issuance protocol that is Lightning compatible would enable PKT Cash liquidity and ultimately be a building block in the creation of both a PKT trading market and in a bandwidth leasing market that is ultimately goals of the PKT network.
* 'scope': 0.5, At the current estimated value of PKT, each milestone costs about $30k. At 4 milestones this is  $133,333, but no timeline is given and no team is yet speculated. My experience with Caleb is that he is fair and honest,  but it is difficult to know for certain what the value of the project is without those elements.
* 'risk': 0.9, Caleb has laid out the protocol's states and messaging, and as the inventor of PKT he knows what he's working with. I believe he has everything he needs to execute this project.
* 'hazard': 0.3 I am involved with Caleb a venture, Anode, which puts me in a conflict of interest in voting for him. However, he is the brilliant inventor of PKT and cjdns. I believe his pedigree and follow-through are self evident. No one knows these technologies better than him.

## Reviewer 2
Tokenstrike seems to be important to me as a long-term goal and I am reasonably
happy with the proposal. Aware that there is the possibility of lack of funding
this time to satisfy all projects, I would be very interested in seeing this
funded at some point, even if not straight away.
* 'short': 0.1,
* 'long': 0.9,
* 'scope': 0.6,
* 'risk': 0.9,
* 'hazard': 0.5

## Reviewer 3
(No commentary)
* 'short': 0.6,
* 'long': 0.7,
* 'scope': 0.8,
* 'risk': 0.9,
* 'hazard': 0.7

## Reviewer 4
* "short": 0.5, Tokens themselves aren't particularly useful, it's the services built on them that matter, so by itself I think this probably doesn't have much short-term impact (but it has *some*, mostly by opening the door for other things to be built on top of it)
* "long": 1.0, Long term this is strictly necessary for the project goals (bandwidth market, etc.)
* "scope": 1.0, I can't think of anything out-of-scope about this. Since we only have enough budget for 1 project, all of them effectively have the same opportunity cost, but this one has the lowest real cost (e.g. we could open another submission window again sooner than if we went with another proposal)
* "risk": 0.75, It's plausible that the loosened security requirements (that attacks can be proven to have taken place, instead of being blocked outright) and a few implementation details could mean that the result is insufficient for some intended use cases. That said, I would naively expect it to be good enough for a large enough number of cases that it's likely to still be useful, and if it was unfit for an interesting reason then at least we'd learn something from it.
* "hazard": 0.75, The proposal was submitted by an NS member, so there's some appearance of nepotism (but realistically, this isn't something that I'd feel confident in supporting without CJD's oversight, so that partly cancels out).

## Reviewer 5
* 'short': 0.8, mechanically set to the average of the other two projects because I have a confilct of interest
* 'long': 0.45, mechanically set to the average of the other two projects because I have a confilct of interest
* 'scope': 0.45, mechanically set to the average of the other two projects because I have a confilct of interest
* 'risk': 0.75, mechanically set to the average of the other two projects because I have a confilct of interest
* 'hazard': 0.4, mechanically set to the average of the other two projects because I have a confilct of interest

PR #35, opened 2020-07-25
Submit 2020_07_24_Ongoing_Tech_Support.md Proposal

johnsonjh:
Greetings, 

I (*Jeffrey H. Johnson*), am submitting this proposal on behalf of Gridfinity, LLC., for review by the PKT Project Network Steward. We (*Gridfinity, LLC.*, *Jeffrey H. Johnson*, *et al*.) are prepared to consider feedback and comments which may be offered by the Network Steward, PKT Project Team, or other stakeholders, before the proposal deadline.

Thank you for your time and attention,


projects/2020_07_25_Ongoing_Tech_Support.md
```diff
@@ -0,0 +1,139 @@
...
+* Projected effort: at least 1 person per month = 24 person-months
+* Requested PKT contribution: 2 million PKT per month x 24 months = TOTAL 48 million PKT
+* PKT address to pay to `pkt1quem94ynpy6mnrgjn7jxc3rq7asp22a5at3fa93`
+* Project status: **PROPOSED**
+
+>
+
+## Project summary
+Gridfinity has been helping the PKT network with critical tasks required to maintain and keep the blockchain operating properly for over a year. The present project proposes that the Network Steward provides support for the expenses of the ongoing technical services and associated administrative tasks that will continue to be provided by Gridfinity to the PKT project. We are not seeking any retroactive payment, but we wish the Network Steward's support to enable us to continue providing the same level of technical services and support moving forward.
+
```
backupbrain, line 19: I've been on many calls with Gridfinity that have revolved entirely around fixing bugs in PKT, the the totality of which have resulted in major feature and stability improvements in PKT.

I applaud Gridfinity's dedication and diligence, and I think this proposal is an excellent idea for them to be compensated for their work in making PKT a better tool for everyone.

projects/2020_07_25_Ongoing_Tech_Support.md
```diff
@@ -0,0 +1,139 @@
...
+Payments will be distributed to Gridfinity monthly, or, alternately, a different schedule negotiated with the Network Steward.
+
+
+
+## 5. Disclosure
+We hereby submit this application in good faith and we attest that we have made no effort, nor do we intend to make effort, influence the Network Steward to accept this or any other project we have submitted.
+
+1. Conflicts
+  1. [x] An organization is receiving the funds: *Gridfinity, LLC.*
+    1. [x] Organization has financial relationships with one or more reviewers: *Caleb James DeLisle* and *Adonis Gaitatzis*
```
backupbrain: This relationship is true, but also Gridfinity has put in over a year of in-kind work into PKT and regardless of my relationship with Gridfinity or its members I believe their work speaks for itself in this regard.

cjdelisle:
Thank you for your proposal. Unfortunately it did not succeed, of the reviewers I was the only one who outright failed the project, the reason why I did this is because the project's long duration and the inherent price instability of crypto could lead to a situation where the currently reasonable salary request grows to a scandalous number and as the project lacks clearly defined success criteria, the NS would not even be able to fend off accusations of corruption by pointing to exact deliverables. The other reviewers were much more positive about the project but their votes were not enough to make it win.

For your convenience, I have copied the evaluations from the [script](https://github.com/pkt-cash/ns-projects/blob/master/vote_2020_06_25.py) where they were written to this comment.

## Reviewer 1
I believe this project will result in dramatically improved reliability, speed, and usability of the  PKT blockchain in the near term.
* 'short': 0.9, I gurantee this project will result in immediate, tangible improvements to PKT and the chain. As party to discussions with the Gridfinity team on existing issues, I know they are already commited to improving and stabilizing PKT and have already put in considerable successful effort to do so at their own expense.
* 'long': 0.1, The improvements Gridfinity seeks to make are incremental and not revolutionary.  They will improve the stability and speed of the chain but are unlikely to revolutionize how PKT works.
* 'scope': 0.7, Gridfinity is asking for 25000 PKT/hour, which at it's current estimated value is around $83/hour, and about 2 days per week dedicated to PKT improvements both sounds reasonable and in line with what I believe to be the time they have put into PKT in the past. In thir proposal, they estimate 2m PKT/month but without an end date. Understandibly, this is a maintenance proposal and therefore it is difficult to estimate the timeline. Bugs are squashed as they are found. But a major limitation of this proposal is  that there is no difinitive end and therefore no way to calculate the total cost in accordance to our budget.
* 'risk': 0.9, Gridfinity has been working with Caleb to improve PKT for over a year. They rely on PKT for their business model. Their developer, Jeff has make specific improvements and bug fixes to PKT in the past and knows the code-base. Therefore I believe they have the capability to do this work and are motivated to do so.
* 'hazard': 0.3 I am involved with one of Gridfinity's partners, Jon in another venture, Anode. This puts me in a conflict of interest in voting for this team, but I believe that fact is mitigated by the pedigree of their past work regarding PKT.
 
## Reviewer 2
Gridfinity right now is pretty central to the network functioning and overall
their input and support has been critical, even if it's just in the form of
fixing bugs and providing technical know-how. This is the kind of support that
the project needs at this stage, although it should be noted that the NS can
not continue to fund projects like this indefinitely without clearer end goals 
* 'short': 0.8,
* 'long': 0.6,
* 'scope': 0.7,
* 'risk': 0.9,
* 'hazard': 0.5

## Reviewer 3
(No commentary)
* 'short': 0.8
* 'long': 0.5
* 'scope': 0.7
* 'risk': 0.9
* 'hazard': 0.7

## Reviewer 4
* "short": 0.95, Tech support is useful, though I tend to think it's not quite as important as core blockchain development is (e.g. there's less there to support before things are feature complete, so better to prioritize developments first and support later, at least in my train of thought)
* "long": 0.5, No direct benefit after the project duration, but presumably some indirect benefit from a larger and more skilled community
* "scope": 0.83, (40/48, rounded, for lack of a better idea, since it's kind of hard to compare cost-per-person-month for focused development work in the other projects vs something as open-ended and wildly variable tech support). I think that, given the budget constraints in the current proposal period, asking for 2 years is a bit much. I think it would be inappropriate to renegotiate the duration after we've already voted, but for future reference, it probably would have made sense to update this proposal with an option to approve it for a shorter duration (1 or 1.5 years) at a lower budget. Then we'd effectively be choosing between tokens+support vs the OT wallet, instead of only getting to pick one.
* "risk": 0.95, I can't really think of a failure mode for tech support, so it seems unlikely that there would be any problems here (at least anything unique to this proposal)
* "hazard": 0.5, This is a little hazardous for 2 reasons. 1: The applicant(s) are known by several NS members, so there's the usual risk of some appearance of nepotism. 2: It seems like, given that Gridfinity both operates a large mining pool and sells mining hardware, they're in a position to financially benefit more from the existence of competent tech support (to increase the size of the community by both selling more hardware and likely picking up more miners through being the default pool in most documentation). So we'd basically be paying Gridfinity to do work that mostly benefits Gridfinity. On the other hand, *no* tech support doesn't really benefit anyone, so I'd begrudgingly consider this acceptable given the lack of alternatives.

## Reviewer 5
* 'short': 0.8, The idea of having someone spend time patching and updating pktd definitely has short term benefit to the project
* 'long': 0.6, My feeling is that longer term, as more pools and miners establish themselves, this will become less and less of a necessity
* 'scope': 0.3, Under a normal cost analysis right now, the project is reasonably priced, but because it is stretched over such a long period, 2 years, there is a distinct possibility for the applicant to end up getting paid an exorbitant amount for their work. Not that any other applicant is getting less, but other applicants are expending labor *now*, at a time when the price of PKT is very low. If the timeframe is in the 3-6 month range then I would give this an 8.
* 'risk': 0.9, There is little risk of the project not succeeding because the project lacks solid metrics to define success
* 'hazard': 0, I'm going to fail this project on the hazard scale because of the combination of the long timespan and lack of success criteria. While I don't believe it is the intent of the applicant, I find the success criteria to be lax enough that the project could be successful without any work done other than the day-to-day operation of a for-profit mining pool. This hazard is exasperated by the timespan of the proposal. In one year's time, the now reasonably payment could become so exorbitant as to bring scandal to the institution of the NS, and without clearly defined written success criteria, we would be unable to prove to the incredulous examiner that this was anything more than a handout to friends of the NS. If the timespan were 3 months and there were clearly defined success criteria then I would have nothing to say against this project but as it is I cannot accept it.

PR #36, opened 2020-07-25
2020_07_25_OTRune_wallet.md

mehowtv:


benhylau:
This PR deletes `projects/template.md`. It shouldn't.

cjdelisle:
Thank you for your proposal! Unfortunately it was not accepted in this round but I hope to see slight variations on it re-submitted soon. The network steward budget in this round was 80mn PKT and [the process](https://github.com/pkt-cash/ns-projects/blob/master/acceptance_process.md) is to rank and collect projects until 90% of the budget is committed. This is performed automatically by a [python script](https://github.com/pkt-cash/ns-projects/blob/master/vote_2020_06_25.py) and I want to attract your attention to the fact that a project of the same quality but half the size would have been funded, so I urge you in the future to consider splitting the work over multiple projects.

As per [the process](https://github.com/pkt-cash/ns-projects/blob/master/acceptance_process.md), the reviewers evaluate project proposals with regard to 5 specific criteria. You can read all of the reviews in the python script linked above, but for your convenience I have places the reviews here as well.

## Reviewer 1
I think this project is very interesting because it is both a near-term need (UI wallet) and has some interesting technological improvements to PKT that would increase PKT's usability.
* 'short': 0.8, I believe that having a desktop UI is critical for widespread consumer adoption of PKT
* 'long': 0.4, Though a UI walletw will be a long-lasting part PKT, I believe the off-chain and decentralized nature of the wallet has important utility for PKT future
* 'scope': 0.8, The timeline and budget seem reasonable for this project, given my understanding of the scope and talent involved.
* 'risk': 0.95, The team seems to have excellent pedigree, including having developed the BIP-47 protocol
* 'hazard': 0.8 There is already a traditional wallet being developed, but without BIP-47 support.  I would prefer that these two projects collaborate, to prevent segmentation of users and hopefully to leverage existing and new tech into a faster shipping time and for the developers involved to all benefit from understanding the latest tech.

## Reviewer 2
Feels to me like pushing a second wallet with different functionality in the
early stages of the project may result in some confusion and fragmentation,
although I do definitely understand the appetite for some diversity long-term
and certainly lightning-like transactions will be beneficial in the future. 
* 'short': 0.2,
* 'long': 0.6,
* 'scope': 0.7,
* 'risk': 0.5,
* 'hazard': 0.5

## Reviewer 3
(No commentary)
* 'short': 0.7
* 'long': 0.7
* 'scope': 0.5
* 'risk': 0.8
* 'hazard': 0.9

## Reviewer 4
* "short": 0.95, Not much to say, if it does what it says on the tin then I'd expect it to have a significant impact short-term
* "long": 0.5, It helps, but wallets are the sort of thing that I expect would emerge somewhere in the community if given enough time and continue to gain features ~indefinitely, so in the grand scheme of things I would expect that accepting this proposal would have diminishing implications over time
* "scope": 0.5, The wallet seems generally in-scope, the main issue is the opportunity cost -- this would take our entire budget for this proposal period, and we wouldn't have anything left over to let us open the next proposal period early
* "risk": 0.5, Two potential issues. 1: I don't know the applicants (at least the ones listed on the proposal, as far as I'm aware), so there's at least a little uncertainty about how prepared they are to work on this, though their related work is generally encouraging (so this has a minor effect -- if anything it's probably good, since it means there's no nepotism hazard). 2: the more significant issue -- the "planning fallacy" -- is that the proposal aims for a 3(+) month schedule with a dozen people working. I'm happy to see the + in the initial outline, but I think *in general* people have trouble predicting how long it will take to do something (especially when other people are involved, a dozen+ in this case). The milestones do a pretty good job of outlining what the applicants want to have happen each month, but I think it would benefit a bit from documenting which components depend upon each other, which parts of the milestone(s) could reasonably be delayed without significantly altering the project timeline as a whole, and what (if anything) can be done to mitigate any issues that arise if a critical component takes longer than expected. Basically, I'd like to hear what a worst-realistic-case scenario sounds like, since (statistically speaking) timelines tend to advance a little worse than that on average
* "hazard": 1.0, I'm not aware of any conflicts of interest or other issues that would lead to problems here

## Reviewer 5
* 'short': 0.8 This is a very impressive set of features to be developed, while it will not distinguish PKT from other blockchains, this will provide a very solid path forward for the project.
* 'long': 0.3 The Open Transactions system can potentially be a good answer to the need for micropayments for a bandwidth market and tokens for representing bandwidth resources, however these solutions are also being developed by Lightning Network developers and many Ethereum-like smart contracting platforms and this proposal does not seek to address any of the more niche requirements of the PKT project.
* 'scope': 0.6 There are two ways that this proposal's success criteria can be read. If we ignore the line "Perform all the already implemented OT transactions that are available on a Notary on PKT, BCH and BTC." then what we have is a wallet with off-chain transacting capability and colored coins which supports iOS, Android, Mac and PC. If we consider that the above line means support for the above-documented "Transfers", "Cash", "Cheque", "Voucher", "Invoice", "Market Offer", "Recurring Payments", and "Smart Contract", then we have additional features, including an important feature which is built-in decentralized exchange functionality. Unfortunately we cannot afford a charitable reading of a proposal when later on we must decide whether the delivered work warrants payment. However, a wallet supporting the major operating systems with colored coin and off-chain transaction support is still a significant amount of work. It is hard to quantify exactly how much work is needed to produce these results, clearly a basic wallet can be achieved for cheaper, for example by a simple Electrum fork. However, it is clear that the Open Transactions team has historically taken the high road and invested significant effort into building their technology and there is no reason in my mind to think that the person-month figure is a scam. The person-month cost is 1.78mn PKT per person-month, making this the cheapest project per declared person-month. However we must also consider the value to the PKT project. As I had said before, a good set of the real needs of the PKT project can be more cheaply achieved with different technologies, so while the work-effort is an incredible bargain, the results are competitive but not outstanding.
* 'risk': 0.6 The success criteria for this project are very well specified, payment is evenly spaced between the milestones and even if milestone 1 is the only milestone successfully delivered, it contains a nice set of success criteria which constitutes delivery of a working desktop wallet app. These risk controls are of very high quality though it must be recognized that the budget and projected effort of this project is larger than any project which has come before it and therefore the inherent risk of the project is particularly high. The up-front payment is 5mn PKT which is not insignificant but is small in comparison to the scope of the project as a whole.
* 'hazard': 0.8 As far as I am aware, there are no conflicts of interest with the NS team. I also see no obvious sign that the project benefits one party over any other. There is a 5mn PKT up-front payment which means the applicant would benefit from running with the loot, but this applicant carries with them a fairly reasonable reputation which would be lost if they were to do so - put differently, they would have to believe that PKT has no future in order for that move to be advantageous. The timeframe given is "3+ months" which is unclear about exactly how much time the applicant plans to spend before submitting milestones, one form of arbitrage which the NS is attentive to is using a project as an "option" to do the work and acquire the PKT later on - pending a rise in the price, but to abandon the project otherwise. It does not seem that this is the case here. Another form of arbitrage which must be considered is whether the amount of work required is in fact 43 person-months, this is a more complex question because it is difficult to ascertain how much work must be done to achieve the stated milestones. Because, as stated in section 3 "Use of Resources", there are likely faster ways to achieve similar objective but there is no evidence of over-charging, I find that a perfect score cannot be awarded but I do not wish to deduct too many points for what is simply a complex topic - opaque to evaluation but by no fault of the applicant.

PR #47, opened 2020-10-02
Milestone 2 report

backupbrain:
Milestone 2 Report for VPN utility

### DIFF
--- a/projects/2020_02_28_pkt_vpn_utility.md
+++ b/projects/2020_02_28_pkt_vpn_utility.md
@@ -736,6 +736,121 @@ A [landing rock for the Anode website](https://anode.co/) has been created, whic
 ### Conclusion
 The project is currently ahead of schedule in Milestone 1, having delivered more than specified in the proposal. We expect to proceed into Milestone 2 without any complications. The error-reporting is allowing the development team to more easily isolate and resolve potential problems with the app and its integrations.
 
+
+
+## Milestone 2 Report - 2020-05-21
+
+### Overview
+
+This document outlines the state of progress through the completion of the milestones outlined in the approved project proposal.
+
+### Progress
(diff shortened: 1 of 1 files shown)


cjdelisle (approved):
This milestone has a few parts missing so it could be reasonably argued that it has not been successful, in the original specification was: Server site login, Client site login, App push notifications, App login with pin, which have been deferred.
However registration of a VPN service and cjdns network access (from M3) have been fast-tracked and ended up in M2.
While I acknowledge my own involvement in the project constitutes a conflict, I personally think this is acceptable because the project has shown real material progress and the applicant seems to be committed to completing the project and exploiting the result.

benhylau (approved):

neilalexander (changes requested):
I'm impressed to see that quite a bit of functionality has been brought forward from later milestones which has led to some significant progress being made. The fact that the core functionality of the app works is excellent news.

However, I am aware that there are quite a few action points from the original proposal that have been deferred. The milestone report is the perfect place to outline where changes to the original plan have to be made and I would request that the report be updated to outline how these items will be addressed and the orders/estimated timescales anticipated. We will then have the relevant history and context to evaluate deliverables when it comes to reviewing the next milestone report.

backupbrain:
@neilalexander I like this feedback.  

It's true we deferred some features as we focused on connectivity first instead of the original plan to focus frameworks first.

The other major deviation is that many features outlined in the original plan, including email/sms communication and certain account management features became obsolete once we used cjdns as the authorization mechanism.  This change benefitted us by allowing us to let users connect to a VPN anonymously, without having to provide a name, email, or phone.

cjdelisle:
@backupbrain please propose an update to the PR to address @neilalexander 's concerns. I don't want to sign it without his concerns addressed.

PR #51, opened 2020-11-14
Create 2020_11_14-pkt-community.md

josh-berger:
Please review our proposal to help manage and educate the ever-growing PKT community.

Thank you for your time and consideration.

Respectfully,
Josh

### DIFF
--- a/projects/2020_11_14-pkt-community.md
+++ b/projects/2020_11_14-pkt-community.md
@@ -61,15 +61,6 @@ Henry Fong is an experienced community builder with experience in branding, onli
 
 **NOTE**: All creative results of the project will be licensed under the [Creative Commons Attribution 4.0](https://creativecommons.org/licenses/by/4.0/) license.
 
-## Success criteria
-
-These success criteria are based on the theory that there are many people in the world who *do* feel disenfranchised by monopoly ISPs and that those people would be excited about the project, and mesh networking overall, but they are currently unaware of it.
-
-* Completion of all assets
-* S_CHAT: 10x growth in number of people registered on pkt.chat - currently 47
-* S_VISITORS: 10x growth in number of people visiting pkt.cash - currently 70/day
-* S_MINING: 10x growth in number of people mining PKT - currently 150
-
 ## Milestones
 
 ### Milestone 0 (Kickoff)
@@ -88,8 +79,6 @@ After the project is accepted, the Network Steward will pay **12 million** PKT
 * WWW_BLOG - launched
 * CHAT_MOD - Code of conduct written and adopted
 * WWW_UPDATE - Update wallet setup instruction and link to blog and docs and to pkt.chat
-* S_VISITORS - 10x growth in pkt.cash visitors
-* S_CHAT - 10x growth in pkt.chat participants
 
 #### Payout
 After a report is delivered on this milestone and it is approved by the NS, the NS will send **8 million** PKT.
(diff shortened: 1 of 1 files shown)


Arceliar:
Hi @josh-berger,

After a brief discussion among the NS, we concluded that it would be more appropriate for this proposal to remove the 10x growth targets for daily site visitors to pkt.cash, registered users on pkt.chat, and miners, from the success criteria for the project and the milestones. Content-wise, I believe this would consist of removing L69-71, L91-92, and L107 of the initially submitted proposal, though you may wish to go over the main text again in case anything is worded in a way that no longer makes sense without those success criteria (unofficially speaking, nothing stands out, but it wouldn't hurt to take a look). In effect, this means we would deliver the milestone awards based on the project deliverables themselves, without being contingent on meeting the milestone growth targets.

Feel free to stop reading here. Since I'm the one who proposed these changes, in the interest of transparency, and to document this for future proposals, I think I should explain my concerns. Lets imagine we approved the initial draft of the proposal, all milestone success criteria were met on schedule, and payments happened exactly as outlined in the proposal. Then in effect, the following sequence of events has taken place:

1. We accepted your "investment" in PKT by accepting a "thing of value" (proposal deliverables) in exchange for the potential for future earnings (the PKT we payed out as milestone awards).

2. The thing we are paying you for is to increase the size of the community (10x miners, site visits, and chat users). The project deliverables exist to achieve those growth targets, so if the growth targets are met, then it's hard to claim that we're not satisfied with whatever you delivered. Likewise, if the growth targets are not met, then it's hard to claim that we're satisfied with what was delivered, so the milestones essentially reduce to the growth targets being met.

3. Some of the people who are brought into the community as part of the milestone targets are giving us a "thing of value" (the various forms of community involvement, which we demonstrably value because your pay was contingent upon facilitating it) in exchange for the potential for future earning (PKT from mining, the statistical near certainty that some fraction of non-miners will purchase PKT at some point, or from their own project proposals accepted by the NS, or whatever other means of acquiring PKT emerge as part future projects). That means at least some fraction of the new community members are also "investors" in the broad sense.

So you would be "investing" in PKT by doing some form of work, but the work we're compensating you for is explicitly the act of bringing more "investors" into PKT. By funding this, the NS's role in operating PKT would start to look an awful lot like a pyramid scheme or possibly a Ponzi scheme (depending on exactly what the new community members are "investing" with -- work from NS proposals, mining effort, or by purchasing PKT from people who already have some). It's not exactly the same, but it's close enough that the NS's actions start to become problematic.

So long story short, we concluded that removing the 10x growth targets is the best way to alleviate this concern. This removes the *explicit* incentive for you to increase the size of the community, and moves much of whatever risk is associated with the project from you (whether or not you can meet your growth targets to receive milestone awards) onto the NS (whether or not the work we have given the award for is enough to meet our growth targets). By assuming the additional risk, the actions of the NS take on a less predatory appearance, which I believe is important for maintaining the legitimacy of the NS and for the long-term health of PKT.

-- Arc, on behalf of the Network Steward

josh-berger:
That's fine. I have made those changes. We will use this project for create of materials and community management.

neilalexander (approved):
Overall I am much happier with the proposal now that it has been amended as requested. I'm very aware that the website and promotional materials have a fairly unique impact on first impressions and how the project is perceived, so I am keen to see more progress on that front. These are the "face" of PKT at present, so it is important to get right. I've been encouraged by the work so far.

However, there's a line of caution between outlining success criteria that produce tangible deliverables versus success criteria that promise certain growth and/or investment, and our approving on the basis of the latter could be misconstrued, so this needs to be considered when submitting proposals.

benhylau (approved):

cjdelisle:
This was accepted in the vote made here: https://github.com/pkt-cash/ns-projects/blob/master/vote_2020_11_08.py

PR #52, opened 2020-11-14
Create 2020_11_14-double_wallet.md

jesseberger00:
This is a proposal for two (2) PKT wallets to be developed and released open source.

### DIFF
--- a/projects/2020_11_14-double_wallet.md
+++ b/projects/2020_11_14-double_wallet.md
@@ -1,0 +1,125 @@
+## double_wallet project
+
+* Project Name: double_wallet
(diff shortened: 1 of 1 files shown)


neilalexander (approved):
I am encouraged by the idea of exploring more than one approach to off-chain transactions as a way of hopefully improving chances of success, but if nothing else then this project will deliver more than one functioning PKT wallet which has obvious long-term benefits.

benhylau (approved):

cjdelisle:
This was accepted in the vote made here: https://github.com/pkt-cash/ns-projects/blob/master/vote_2020_11_08.py

PR #57, opened 2020-11-21
changes as per @vneilalexander request

backupbrain:
I've explained a little why we've deferred some deliverables from earlier milestones, and also explained which deliverables are obsolete as a result of a change in how we handle authorizations (now with cjdns  public keys without the need for email/password).


neilalexander (approved):
Thanks for the clarifications!

cjdelisle:
Accepting because this is a clarification of #47 which was merged and this answered the issues which were raised in it.

PR #70, opened 2021-04-27
Perhaps it's about time we opened a call ?

cjdelisle:
What do you think ?

### DIFF
--- a/projects/template.md
+++ b/projects/template.md
@@ -132,6 +132,28 @@ submitted.
     * If selected, please attach a paragraph detailing the information which will be presented and any steps which will be taken to prevent this from potentially misleading the public.
   2. [ ] Project results will not present information which might lead to PKT price speculation
 
+## Use of Resources
+Each milestone report will be accompanied by:
+
+* [ ] Itemization of project expenditures, including description, date and price (in national currency)
+  * [ ] Justification of this itemization in an audit certificate
+* [ ] Dates when PKT was converted to national currency
+  * [ ] Justification of this conversion in an audit certificate
+* [ ] An audit certificate from a qualified financial auditor
+  * *If yes, please provide name of the auditor*: ___________
+
+> If there is any other information which you can agree to provide that will help the network
+steward validate your use of resources, please add it here.
+
+### Partial milestones
+If it happens that I have satisfied all of the criteria for a milestone but have not provided
+justification of all resources allocated *(choose one)*:
+
+* [ ] I will deduct the un-justified amount from the amount requested for the milestone
+* [ ] I will request the milestone be paid in full anyway
+
+> NOTE: The network steward might reject a milestone or offer partial payment in any case.
+
 ## Legal
 
 The applicant understands that the network steward is not a legal entity and no part of this
--- a/readme.md
+++ b/readme.md
@@ -4,7 +4,7 @@
 * **CURRENT ROUND BUDGET**: 40mn PKT
 
 ## IMPORTANT RECOMMENDATIONS
-Following a conversaion between the members of the Network Steward team, it has been decided that
+Following a conversation between the members of the Network Steward team, it has been decided that
 the Network Steward shall make the following **recommendations** to applicants:
 
 1. Applicants are recommended to provide in their applications, a commitment to including, in
@@ -162,4 +162,3 @@ to submit the project to this repository
 If you submit early then the Network Steward team might provide feedback but others can also see
 your project proposal. If you submit at the deadline then the project will be secret up until
 the end, but it will be evaluated as it is.
-
--- a/readme.md
+++ b/readme.md
@@ -1,7 +1,31 @@
 # Projects Supported by The Network Steward
 
-* **CURRENT DEADLINE**: NO round currently active
-* **CURRENT ROUND BUDGET**: -
+* **CURRENT DEADLINE**: Midnight UTC - May 22nd 2021
+* **CURRENT ROUND BUDGET**: 40mn PKT
+
+## IMPORTANT RECOMMENDATIONS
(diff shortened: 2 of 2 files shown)


neilalexander (approved):

readme.md
```diff
@@ -1,7 +1,31 @@
...
+above statements.
+3. Applicants are recommended to specify in their applications that they agree to receive partial
+payout for a milestone if they can only justify partial use of the resources from the previous payment.
+
+**Please Note**: These recommendations are not prescriptive, the Network Steward wants to verify that
+resources were allocated to the execution of projects, even if purchasing power may have changed as a
+result of PKT price instability. It is the job of the applicant to describe the process which they
+intend to follow, but the Network Steward recommends that applicants offer as much detail as possible,
+including such things as the name of their accountant and which exact statements the audit certificates
+will support.
```
Arceliar, line 28: I'm not a big fan of the wording in this last note paragraph. It may be something to revise after the next review period, when we've received feedback from the applicants.

Arceliar (approved):
Looks good to me, no major comments or concerns. I'm not 100% happy with the wording in the last "note" paragraph, but I think it's good enough for now.

projects/template.md
```diff
@@ -132,6 +132,28 @@ submitted.
     * If selected, please attach a paragraph detailing the information which will be presented and any steps which will be taken to prevent this from potentially misleading the public.
   2. [ ] Project results will not present information which might lead to PKT price speculation
 
+## Use of Resources
+Each milestone report will be accompanied by:
+
+* [ ] Itemization of project expenditures, including description, date and price (in national currency)
+  * [ ] Justification of this itemization in an audit certificate
+* [ ] Dates when PKT was converted to national currency
```
backupbrain, line 140: I think this is super important. Otherwise the NS could be at risk of holding on to PKT until it's worth more, even if it strings along the developers.

projects/template.md
```diff
@@ -132,6 +132,28 @@ submitted.
...
+  * *If yes, please provide name of the auditor*: ___________
+
+> If there is any other information which you can agree to provide that will help the network
+steward validate your use of resources, please add it here.
+
+### Partial milestones
+If it happens that I have satisfied all of the criteria for a milestone but have not provided
+justification of all resources allocated *(choose one)*:
+
+* [ ] I will deduct the un-justified amount from the amount requested for the milestone
```
backupbrain, line 152: It would be interesting to have a bonus mechanism, maybe a default 5 or 10% withholding that gets paid out when milestones are completed ahead of schedule, so as to incentivize teams to develop faster.

projects/template.md
```diff
@@ -132,6 +132,28 @@ submitted.
...
 
+## Use of Resources
+Each milestone report will be accompanied by:
+
+* [ ] Itemization of project expenditures, including description, date and price (in national currency)
+  * [ ] Justification of this itemization in an audit certificate
+* [ ] Dates when PKT was converted to national currency
+  * [ ] Justification of this conversion in an audit certificate
+* [ ] An audit certificate from a qualified financial auditor
+  * *If yes, please provide name of the auditor*: ___________
```
benhylau, line 143: Do we need more info such as contact info?

readme.md
```diff
@@ -1,7 +1,31 @@
 # Projects Supported by The Network Steward
 
-* **CURRENT DEADLINE**: NO round currently active
-* **CURRENT ROUND BUDGET**: -
+* **CURRENT DEADLINE**: Midnight UTC - May 22nd 2021
+* **CURRENT ROUND BUDGET**: 40mn PKT
+
+## IMPORTANT RECOMMENDATIONS
+Following a conversaion between the members of the Network Steward team, it has been decided that
```
benhylau, line 7:
```suggestion
Following a conversation between the members of the Network Steward team, it has been decided that
```

readme.md
```diff
@@ -1,7 +1,31 @@
...
+the Network Steward shall make the following **recommendations** to applicants:
+
+1. Applicants are recommended to provide in their applications, a commitment to including, in
```
benhylau, line 10: What happens if developers are compensated in PKT? I imagine it'd be booked as "equivalent $" like any other project expenditure right? If so, we should probably mention this somewhere.

projects/template.md
```diff
@@ -132,6 +132,28 @@ submitted.
...
+
+> If there is any other information which you can agree to provide that will help the network
+steward validate your use of resources, please add it here.
+
+### Partial milestones
```
benhylau, line 148: Lots of the indentation is messed up, perhaps this is an opportunity to fix them all?
See: https://github.com/pkt-cash/ns-projects/blob/073fe3662e67a8c0dfbb0abd97cd2a070a19e752/projects/template.md#disclosure

benhylau (approved):
Approving content, but pls see comments before merge.

projects/template.md
cjdelisle: Can you make another PR to do that so we're not blocking acceptance of opening a new round on it ?

readme.md
cjdelisle: See: https://github.com/pkt-cash/ns-projects/pull/70/files#diff-5a831ea67cf5cf8703b0de46901ab25bd191f56b320053be9332d9a3b0d01d15R15
Projects will be PKT denominated as before, no change, this is just recommendations for them to do accounting.

projects/template.md
benhylau: Feel free to make changes in another future PR, and accept this as is.

readme.md
benhylau: Yes I get that they don't have to be sold, just thinking to make explicit mention of wage expenses. Ok to merge as is tho.

cjdelisle:
Looks like we're a go

PR #71, opened 2021-05-08
User Operated Internet fund - Request for Guidance

cjdelisle:
Hello folks,
I wanted to propose this project as a way to lift the load off of the Network Steward shoulders. I know we all have felt the pressure of the future looking back on us as we try to do our best to allocate resources for the benefit of the ecosystem, and I really appreciate the seriousness which everyone has put on this role. I have put this project together to move the heavy-lifting operations of the Network Steward to a public benefit charity with a legal infrastructure and professional full-time reviewers, so that the heavy lifting can be done by professionals while the oversight remains with the NS team.

I am putting in this project as a Request for Guidance, which means I want your advice:

1. Is this project structurally sound? It has a flexible budget size, is that OK? Is it something that can compete in your mind?
2. We considered using the name "Network Steward Trust" for the Netherlands not-for-profit foundation, is this a fine name or would you prefer we used something different?
3. I put myself as *not* having the ability to influence a project to make it win, yet retain the power of veto in case there is a project which I consider unacceptable / dangerous to be funded from Network Steward resources. Is this a good idea in your mind?
4. This project is meant to serve the NS, is there anything else which you think could make it better?

Thanks,
Caleb

### DIFF
--- a/projects/2021_05_09_user_operated_internet_fund.md
+++ b/projects/2021_05_09_user_operated_internet_fund.md
@@ -1,0 +1,285 @@
+# User Operated Internet Fund
+* Project Name: User Operated Internet Fund
+* Contact Email: cjd@cjdns.fr
+* Project participants:
+  * Caleb James DeLisle
+  * Michiel Leenaars
+  * Bob Goudriaan
+* Projected duration: 8 months + time needed to spend remaining budget
+* Projected effort: Not specified, see accounting section for details
+* Pre-project effort: -
+* Requested PKT contribution: 10mn, 20mn, 30mn, OR 40mn PKT
+  * This project can operate with as little as 10mn PKT, but can effectively deploy as much as 40mn PKT, so we request that the network steward votes this project as though it were 4 separate projects such that other proposals which are deemed better than this project can win and this project will still be able to deploy some resources.
+* PKT address to pay to: pkt1qku865lg0jxkwpl4pvrp8wkzkkjwna4y8ts243v
+
+## Project summary
+
+This project will take PKT into *Stichting Network Steward Trust*, a Netherlands not-for-profit foundation (hereafter called "the Trust") set up especially to manage cryptocurrency assets dedicated to the public benefit. The Trust is open to any donations of cryptocurrencies serving its mission. PKT held in this entity is subgranted to a professional grant fund managed by the [NLnet Foundation](https://nlnet.nl), a recognised and forward-looking philantropic organisation established in 1989 in the Netherlands. The fund provides EUR denominated micro-grants to individuals and teams with projects considered relevant to the PKT ecosystem.
+
+This setup allows to provide and manage much more fine-grained mechanisms that deliver subgrants to the community through e.g. open, competitive calls which are open to individuals and organisations of any type from around the world. A dedicated full-fledged fund with a long term roadmap also allows to arrange for various horizontal support mechanisms to provide quality measures, liaise with the technical and operational internet community, the press as well as create consistency and synergy across a large set of smaller efforts. Being part of a larger whole and receiving tailored support helps boost the maturity of outcomes (through e.g. standardisation in bodies like IETF, IEEE and W3C) as well as improve real-world deployability. Such horizontal support measures includes involving specialised community partners (typically not-for-profits themselves) to provision grant beneficiaries with e.g. security audits, accessibility, packaging, software testing and documentation.
+
+Working through an established philantropical organisation brings along strict regulatory oversight, transparency and professional financial management, and as such delivers strong guarantees regarding avoiding conflict of interest, corruption, and fraud. Dealing with a proper legal entity also provides significantly more legal guarantees to the people involved in projects, given them piece of mind about their efforts.
+
+## Team and Past Work
+
+The following people will comprise the board of stichting Network Steward Trust:
+
+* Caleb James DeLisle - The original author of the PKT software and cjdns, will take a light handed role, he will *not* evaluate projects nor influence the evaluation process, but will be able to veto a project if it is seen as unacceptable use of Network Steward resources.
+
+* [Bob Goudriaan](https://nlnet.nl/people/goudriaan) - General director of NLnet foundation, the philantropic organisation that will perform the grantmaking for the User Operated Internet fund. He is on the advisory board of many different programmes with the Next Generation Internet ecosystem, such as NGI Pointer, NGI DAPSI, NGI LEDGER and TETRA - which jointly manage tens of millions of investment into public benefit R&D. He is a board member of DINL. Since 2014 he has been involved with the ethical security company Radically Open Security, a socially driven enterprise that has meanwhiled donated over half a million euro of its profits to NLnet foundation. Goudriaan studied fiscal law at Leiden University, and afterwards moved back and forth between functions with large organisations like PWC and Deloitte across the globe and entrepreneurial activities. Goudriaan was previously partner and co-founder of the Amsterdam based management consultancy Gendo, specialised in technological innovation. He will be responsible for the financial, regulatory and fiscal management of the User Operatered Internet Fund.
+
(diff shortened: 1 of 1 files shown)


Arceliar:
Regarding @cjdelisle's questions:

1. I'm not theoretically opposed to a flexible budget, but we need to think about how to do that within the limitations of the current voting system (we *could* change the voting system to make it handle this better, but changing it right now could look like favoritism from the NS, so that's probably not a good idea). I would advise highlighting a fixed budget that we *could* use to vote, if we can't figure out a practical and impartial way to include a flexible budget within the existing voting system.

2. I think I'd prefer a name that didn't include "Network Steward" (or prominently feature PKT) to limit confusion between the NS and the trust.

3. I would strongly recommend either giving the veto to the NS as a whole, or keeping the NS members completely out of the per-project approval process. Board membership (even in a veto-only capacity) for a subset of the NS introduces a power imbalance, and in the long term I think that's likely to lead to problems within the NS which outweigh any benefits of direct involvement within the trust. That said, some kind of non-voting (including non-veto) board membership, e.g. as a liaison or in an advisory capacity, is probably a very good idea.

4. Not exactly suggestions, but I have a few questions/comments below.

Can I get some clarification about something in the Bylaws?
Item 3 says it will use a fixed administrative overhead of 10%.
Items 4 and 5 mention up to 15% and 30% of the "entire expended budget" as being for administration and support projects.
Do I understand correctly that the 15% in item 4 is made up of the various 10%s (per project) from item 3, plus any additional (non-project-related) administrative overhead?
So the maximum theoretical overhead (not awarded to projects directly) would be 45% of the total budget, with the rest being given to the funded projects?

Applications from minors:
I'm not sure what the legal risks are for even considering proposals from minors prior to receiving consent from a legal guardian. 
If this were in the US, then pragmatically speaking, I think the legal headaches wouldn't be worth it (I expect minors would submit a relatively small number of applications, and they'd have a below average acceptance rate, so the legal risk and the overhead necessary to maintain compliance seems like it wouldn't be worth it).
I'm not that familiar with the potential legal hurdles in the Netherlands (or more broadly the EU), so I'm not suggesting anything here, I'm just thinking aloud.

The formatting on the "Disclosure" section seems to be a bit messed up when the markdown is rendered.

cjdelisle:
I have pushed a new commit https://github.com/pkt-cash/ns-projects/pull/71/commits/ea4ba795f81167414af8d3b24db83e675eca4902 which addresses:
Point 1: There are 4 projects but the NS should do how it wants
Point 2: Name changed
Point 3: Veto removed, NS can suspend project at any point
Point 4: Multiple clarifications regarding support projects
* Spend nothing except at direction of a "regular" project
* New rule: at least 2/3 of entire budget goes to regular projects in any case
* Application from minors: Clarified, this is standard NLnet practice
* Formatting fixed
* Also minor changes of wording

cjdelisle:
I just pushed a second change which adds a June 1st call. NLnet has a team who are able to work with applicants of good but under-specified projects help them become great projects, and the June 1st call will not only provide a way for the process to start earlier/slower and ramp up, but will also give other projects who may apply to this call but be not quite good enough, a potential second chance / helping hand. I think this change is all positive, but if there is any problem with it then we should be ok to work with the previous version as well.

Arceliar:
The changes help to clarify things on my end. I like adding an additional early call (June 1st seems a little aggressive, but I guess it's OK as a second chance for proposals that come short of NS approval in the current call). I don't think I have any additional comments or suggestions at this time.

cjdelisle:
4 out of the 5 voters have cast their vote. Due to the obvious conflict of interest, the vote which I cast for this project was mechanically set to the average of the other projects which I reviewed. Incidentally this caused me to reject the project. Nonetheless, this project has enough votes to be accepted from the other NS team members.

I have included the evaluations here so that others may use this proposal when structuring their own proposals in the future. As you might know, each vote is made on 5 different items:

* short term value
* long term value
* scope and use of resources (how much value/cost is the project)
* risk (of the project not going well)
* hazard (of real or perceived corruption)

Each evaluator casts a vote which is a number between zero and one. If an evaluator casts a zero vote on scope, risk or hazard, then they are indicating that they find the project unacceptable on that point, also if an evaluator casts a zero vote on both short term AND long term value, then they find the project unacceptable on those points. Below are the reviews which can be used for evolving proposals in the future by yourself and others.

## Ben
I like this proposal and think it is timely, as the NS receives more broadly scoped
proposals that often extend beyond the domain expertise of the NS team. An arrangement
with an experienced and reputatable institution not only takes some load off the NS,
but also adds legitimacy to our ecosystem. I expect a lot of learning to come out of
this engagement.

I recognize the conflict position of groups having financial relationships with Caleb,
who is part of the NS, but understand that clear boundaries and processes are in place
with respect to use of funds. Secondly, this is a big amount to allocate to one project,
but because of the nature of it, we do need a bigger commitment to see results.

Some things I would recommend is, if things don't go well we have an understanding that
the NS can decline to fund the next milestone. By the end of the grant, if the process
goes well, we can revisit the allocation of PKT for administrative vs. grant. Both these
are addressed / possible in the current proposal as written.
* 'short': 0.75,
* 'long': 0.85,
* 'scope': 0.7,
* 'risk': 0.65,
* 'hazard': 0.65

## Adonis

### Short term

I don't believe there will be any short term impact.
I percieve this change is being similar to the splitting of Xorg from
the X11 foundation. In name and principle only and totally without
consequence the time it took for the Xorg foundation to release a new
project.

There's a possibility it will have some impact in that people will
percieve the PKT community as being in support of standards,
best practices and good governance.
* 'short': 0.1,

### Long term
I think this will have tremendous far-reaching consequences for PKT
I believe this will alter the course off what is possible with the
community and the technology, in the same way that the IETF has made
it possible to produce and canonize SPF, SASL, Oauth and more.

Since the payout of this project's fund is denominated in EUR,
I also think that a project like this will reduce the temptation for
clauses such as the one in the lobbying proposal to take a variable
amount of funding depending on the price of the crypto
* 'long': 0.9,

### Scope
This is one of the few projects I can think of that both requires
relatively little upfront work and yet has a lot of long-term impact.
In that way it reminds me of the Network Steward. It took almost
no effort to set up ompared with building a product and yet produces
value to the people who contribute to the PKT ecosystem.

This is something I know personally that CJ has has been talking
about since the start of the Network Steward, so I believe
he will carry it through.

The financing and milestone struture is strange for this projet
It attempts to seek a flexible budget (apparently considering that other
projects deserve funding and so it must be flexible in that funding), and
to consider each payout as a separate project to acheive that, given
the mechanics of funding by the NS.

It's an interesting way to approach flexible funding. I assume the extra
PKT will go to provide more support for projects submitted to it.
However since it lists several support activities it will provide such as
security auditing, and because it can receive donations at any time,
I don't see why it wouldn't seek to create the organization and a smaller
set of services in the beginnig and then ask for more once it's achieved
a smaller goal.
* 'scope': 0.5,

### Risk
One thing I like about this project is that it could bring PKT from
an association of like-minds to a formally recognized body such as the
IETF, complete with oversight that would build public trust.

I would love to fund this alongside Lightman's proposal. I think they
complement each other well.


One thing I saw recently that bothers me is that Signal took a donation
which would have paid for its operation for years, but then got flooded
with new users when Elon Musk tweeted "use signal," which probably reduced
Signal's runway from that donation by eating up infrastructure costs
It happens that Signal was releasing a crypto currency around that time,
which went up 10x in value over the course of a couple weeks, fell 30%
in a day, and then fell back down to its original price over the next few
weeks. It is possible, or it would be at least tempting for them to have
pumped up their own crypto and sold at peak to pay for years of
infrastructure costs.

The NS does not do marketing, so I hope it would be difficult for the NS
to perform a trick like this, but I can envision that the temptation
would be there since not only are some or all of the NS members PKT cash
holders and/or have companies that employ PKT technology and the NS
funds projects in PKT so the budget for funding increases when the price
increases. Having an NGO that can accept donations in any crypto and is
governed by strict government oversight in a country such as the
Netherlands where there is both a favorable tax regime and socially
responsible laws regulating NGOs further reduces this risk.

A potential down-side is that the regulation will make the proposals
and funding process more challenging and time-consuming. In working
with the IETF in the past, I found their feedback process slow and their
submission process technically challenging and antiquated. It may deter
casual / crazy submissions but it also deters great ideas that don't want
to jump through 10 mile hoops.
* 'risk': 0.5,

### Hazard
I work with CJ in both Anode and PktPal, so there may be a conflict
of interest on this. Also CJ is part of the NS, so funding his project
may come off as favoritism.

I personally believe that the vision of this projet is much bigger than
me or CJ, and that CJ has proven his ability to operate this type of
organization through the genesis of the Network Steward.
* 'hazard': 0.5

## Caleb

Because there is a conflict, I have mechanically set the UOI rating to the average of the others, however
I have set hazard to zero instead of one because I have failed both of the other projects so is it best
in keeping with the intent of non-intervension for reason of conflict that I also fail the UOI project.

* 'short': 3/5.0,
* 'long': 2/5.0,
* 'scope': 1/5.0,
* 'risk': 1/5.0,
* 'hazard': 0/5.0,

PR #73, opened 2021-05-22
PKT Wallet GUI with mining from your browser

vantwoutmaarten:


cjdelisle:
Hmm figuring the trading price of PKT to be roughly 0.05 and a 160 hour month (40hrs/wk) this work would be billed at $656.25 per hour.

vantwoutmaarten:
We measure our impact by added value to the PKT community, not by fiat.

cjdelisle:
Understood, that was a note for the NS team when evaluating. As per the rules, we don't communicate our opinions n projects except in public.

vantwoutmaarten:
No problem, thank you for looking into our project!

projects/2021_22_05_pkt_wallet_website_GUI_for_the_blockchain.md
```diff
@@ -127,7 +127,7 @@ __Add the option to announcement mine on from your browser.__
 > * It will be possible to announcement mine from within your account on the website. 
 
 #### Payout
-After a report is delivered on this milestone and it is approved by the NS, the NS will send 4,000,000 PKT.
+After a report is delivered on this milestone and it is approved by the NS, the NS will send 11,000,000 PKT.
```
cjdelisle, line 130: Note that the proposal can only be considered in the form which was proposed as of the time of the deadline for applications.

vantwoutmaarten: Oke I understand, we are looking forward to the results of the current round!

cjdelisle:
Hello, 4 out of 5 of the votes were case and unfortunately your project was not accepted. As you might know, each vote is made on 4 different items:

* short term value
* long term value
* scope and use of resources (how much value/cost is the project)
* risk (of the project not going well)
* hazard (of real or perceived corruption)

Each evaluator casts a vote which is a number between zero and one. If an evaluator casts a zero vote on scope, risk or hazard, then they are indicating that they find the project unacceptable on that point, also if an evaluator casts a zero vote on both short term AND long term value, then they find the project unacceptable on those points. Below are the reviews which can be used for evolving proposals in the future by yourself and others.

## Ben
I appreciate the intention to make onboarding general audience easier, but don't
think that browser-based mining is the best solution. Short-term we may see more
people try it out, but I cannot imagine anyone long-term mining from a browser
tab, even though it is wasm. Then we have to deal with the non-custodial management
of the private key, which if someone knows how to do that, they can manage a native
miner. I am also cautious of the NS funding yet another wallet project at this time,
at this stage of the project it may create more confusion than benefit. I also think
this is both risky and expensive for the potential impact.
* 'short': 0.4,
* 'long': 0.1,
* 'scope': 0.3,
* 'risk': 0.5,
* 'hazard': 0.7

## Adonis
This is a really interesting idea, but frought with risks.

### Short term
I think this project is super interesting. I think one of the issues
with PKT today is that mining is challenging. It requires a Linux computer
a lot of bandwidth, and knowledge of the command line.

I personally have never seen a browser-based miner that works well.
PKT uses a *lot* of CPU and can basically grind a computer to a halt.
Sure, using it in the browser would make it easier to use, but only
for people who are basically willing to give up use of their desktop
or laptop anyway, so in that case why not have a dedicated computer?
* 'short': 0.3,

## Long term
One potential I see for this project is that, since mining is effectively
routing, any home user could became a PKT router, sharing their spare
bandwidth while they are surfing. With that, it may be possible to
encourage normal people to become micro-ISPs for their friends and
neighbors.

I also see potential here for increasing the number of people aware
of PKT tremendously, as deep technical expertise will no longer be
required to interact with PKT. It will be like Bitcoin in 2011-2013,
where people still must be comfortable with keys to send money, but
honestly there are better PKT wallets that people can keep on their
desktop that don't require keeping track of private keys.

But that's all conditional on this being promoted the right way
agree with Arc about the risks of this being percieved as a negative.
This also has the potential to radically polarize people against
PKT as a the symbol for malicious browser-based crypto mining.
* 'long': 0.1,

## Scope
In reading "scope" as "can this project be done?":
I'm not sure why this needs to have a private-key payment system
there are plenty of PKT wallets. I know of 3 already and two more in
development. The miner is interesting and it can send to one of
these wallets.

The other thing is that no wallet tech is listed, so how is the wallet
built? It's difficult to know if this project is even achievable
within scope without knowing if the wallet will be built from scratch
or use an existing tech.

in reading "scope" as "should the NS fund this project?":
I see wallets and miners as generally in scope of the NS, but we already
have so many wallets. Unless one is mind-blowing (and please, show me
wireframes or mock-ups and workflows at this point to prove it), I don't
see the need for yet another wallet.

As a miner I think it's an interesting project, but it carries a lot of
risk. Should the NS fund a miner that could be used maliciously before
PKT is even widely known or used? The potential bad will could outweigh
the good, setting back not only PKT but crypto for decades.
* 'scope': 0.0,

### Risk
If the goal is to build a miner on the browser with the given funds,
probably that's doable. If the wallet is built from scratch, I don't
know how achievable that is. If the wallet is modified from another
project, probably it's doable. It's hard to rate without knowing
what the sallet technology is.
* 'risk': 0.0,

### Hazard
I see several potential issues with this proposal
1) although three participants are described, only one
has a GitHub or website link, despite the team claiming to be composed
'software developers and blockchain and webdevelopment experts'
2) No organization is formed, so the payout could to an individual, but
it's unclear who. This is not a disqualifier per se, but it's unusual
3) a browser-based miner could be used maliciously, so I don't think
the NS should fund a project like this at this point. Perhaps in the
future when the public is more aware of these things or there are
mechanisms to prevent malicious use of this type of technology or
or browsers implement detections to detect malicious blockchain
* 'hazard': 0.0

## Arceliar

### Short term
As far as I know, this would be the first web-based frontend to a PKT
wallet. That at least has some short-term novelty to it.
* 'short': 0.5,

### Long term
There amount of long-term utility is more than none, I guess, but I
definitely see this as having more short term benefit. Worst case
scenario, the code could either be used as a starting point for a new
implementation, or an example of what not to do (if it fails for
interesting reasons).
* 'long': 0.1,

### Scope and use of resources
It's not clear to me if the intent was to write a new wallet, or use an
existing wallet (which?) on the backend. I'll assume the intent is to use
whatever the "best" existing wallet is, since this seems to be more about
the frontend. Either way, this is yet-another-wallet. I don't think it
makes much sense to keep funding wallet projects, since at some point
they'll just be fighting over the same body of users. This rating could
change if the proposal was resubmitted in the future, depending on how the
other wallet projects turn out.
* 'scope': 0.,

### Risk
Assuming this is just a front-end, I don't think this is particularly
risky as wallet-related projects go. If there weren't already other
wallets in the works, then a web-based frontend would probably be one of
the less risky projects to fund, in my opinion. That's not the world we
live in, so I'm not sure the web based wallet would really attract that
many new users (be that from other wallets to itself, or to the PKT
community as a whole). On the whole, I see this proposal as kind of risky,
since it partly works against the other wallet proposals, but not so bad
that it would be disqualifying on its own.
* 'risk': 0.5,

### Hazard
I don't think the NS should fund a web-based miner. I've come across too
many instances of malicious mining code being injected into web pages. If
the NS funded development of a web-based miner, and then it found its way
into e.g. a malicious advertisement, the negative press from that could be
bad for PKT (or at the very least, reflect poorly on the NS). Without that
aspect, I think this would be fine from a hazard aspect (but not great
overall, due to the relative saturation of wallet projects).
* 'hazard': 0.,

## Caleb
Despite significant Network Steward support in the past, the wallet story is clearly not the strongest aspect of the PKT project.
There are currently 3 different wallets, a command line wallet which is robust but not user friendly, an electrum based
wallet which is user friendly but not robust enough to support thousands of payments which occur when mining, and Zulu
which is a GUI frontend for the command line wallet, but which only works on Apple computers.
There is an entirely separate wallet in development by the team who produces OpenTransactions, but this has not yet been released.

My observation is that there is an abundance of wallet "technology" but a lack of "product", so for example technical
support ends up falling on the open source community which means it inevitably has a "helping eachother" aesthetic
rather than one of a smooth experience with a complete product.

### Short term
It is my opinion that the PKT community needs a business to step up and build a wallet product which has a financial model
that puts the customer in charge. The development of additional open source wallet technologies by the Network Steward is
certainly not worthless, but in my evaluation it does not meet the current pressing need.
* 'short': 3/5.0,

### Long term
This project is not described in terms of long term impact to decentralizing internet infrastructure. One might argue that
short term impact drives long term impact, but short term impact has it's own category in this evaluation.
* 'long': 0/5.0,

### Scope
Any endevor to build a crypto wallet must necessarily be based on either existing wallet technology such as PKTWallet
or ElectrumX, or on an entirely new wallet technology. Ideally a proposal such as this would specify whether it is to be
based on new or existing technology, and in the case of new tech to explain why it is important to have this new technology
and how it would work, and in the case of existing technology, to give a brief rundown of the tech stack so that
feasibility, cost, and risk can be evaluated. Since this proposal did not provide any of these details, we need to make a
guess as to intent, with an eye to the minimum deliverable that would satisfy the success criteria.
Since the proposal says "everybody should be able to send and recieve PKT" it seems that the focus is more about bringing
technology to the consumer rather than creating new technology, so I think it is fair to assume that the intent is to
leverage existing technology, which is also the cheaper option.
Building a new wallet based on existing technology should probably take somewhere around 1 person-month of effort to specify
and design, 3 PMs to develop and test and 1 PM to manage, which comes out to 5 PMs total. So the proposal's estimate of 10 PMs
is perhaps high, but not ridiculous.
On cost, the current price of PKT in informal trading seems to be roughly 5 cents, this places the total project cost at
1,050,000 and thus the person-month cost at 105,000.
Assuming every month to be 4 40 hour weeks of fulltime work, the cost would be around 650$/hr which is very high.
I'm not going to outright fail the project on this point but I would only be inclined to accept a project of this cost if
there was nothing better.
* 'scope': 1/5.0,

### Risk
There are 5 milestones and payouts are reasonably spaced. Without knowing what existing wallet technology this project would
be based on, or if new, how the new wallet technology would work, I can't reasonably give this better than 2 of 5.
* 'risk': 2/5.0,

### Hazard
On the question of hazard, the checkboxes for declaration of conflicts were not filled, but I don't know the applicant
and don't know of any conflicts myself. However the applicant is asking for what is, if we consider the informal trading
chat quote of 5 cents, over a million dollar grant. I see no mention of any business entity which would be in charge of the
project so it appears as though an individual would be receiving the funds. It is very irregular for an individual to receive 
a grant of over 1 million dollars in value rather than an entity with a public accountant. Also if the price of PKT were to
rise in the next couple of months, there is no provision in this project for the applicant to receive partial payout for
subsequent milestones which means this project could conceivably become worth tens of millions of dollars and have an
effective person-month rate of tens of thousands of dollars per hour so in my opinion this project fails the test of
"without unfairly benefitting any one participant over any other" and I am going to fail it on hazard control.
* 'hazard': 0/5.0,

PR #74, opened 2021-05-24
Housekeeping: added vote template in order to allow votes on the proj…

cjdelisle:
…ects


cjdelisle:
I'm going to self-merge this so that we can add our votes

PR #75, opened 2021-05-24
Added my vote for the projects in this round

cjdelisle:


cjdelisle:
Going to merge this so we can start working on an actual document

PR #80, opened 2021-06-26
cjdns/wireguard is now delivered in my opinion

cjdelisle:
I just wanted to PR this in because from my perspective the milestone is delivered. I was very much involved in this because it makes a change to the cjdns protocol so I wanted to be sure it was done exactly right.
I did not receive any financial benefit.


Arceliar:
As with an earlier PR, for some reason the button to submit a review request disappears as soon as I log in. Regardless, the dev of the project is saying the submission is satisfactory, so that's good enough for me. OK from my side.

PR #83, opened 2021-11-18
Update and rename template.md to pktapphelper.md

binartech:
pktapphelper

### DIFF
--- a/projects/pktapphelper.md
+++ b/projects/pktapphelper.md
@@ -1,4 +1,4 @@
-# *Project Name*
+# *Pkt miner Helper App*
 
 > **Instructions**: If you are viewing this template in github, click on the pencil icon in the
 upper-right of the pane:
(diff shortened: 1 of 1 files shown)


cjdelisle:
Hello there, thanks for making this proposal !
Could you please not remove then template when adding the proposed project ?

Also we will need some information such as how much PKT you are expecting. If you would like advice then you can join pkt.chat and drop by the #network-steward channel, love to help you make a good project.

cjdelisle:
Hello @binartech , thank you for making this proposal. These are not the opinions of the NS, these are just my personal comments.
Unfortunately this project lacks the level of specificity which I think is necessary for the NS to give it a proper review, such as the expected budget and clearly defined deliverables and success criteria.
I think this type of project might be a good fit for a one-person-software-project, following this template: https://cryptpad.fr/code/#/2/code/view/lHORne-EdOqRVL2kBQ7SCPpt8U9SoRF5utOCFjnaGVs/
You may leave this pull request as it is and it will be officially reviewed by the NS, or you can update it until midnight UTC at the end of March 4th.

cjdelisle:
I regret to inform you that this project was not accepted as it did not contain adequate specificity to be evaluated.

PR #84, opened 2021-11-23
Milestones 1 and 2 for UOI fund

cjdelisle:
Hello all, I'm sorry that this fund has suffered administrative delays, but finally we have been able to fund some very exciting projects and I think you will understand the value of these milestones which we are submitting together.


cjdelisle:
Accepting this as housekeeping because it was signed and paid by the NS

PR #87, opened 2021-12-14
GUIDANCE REQUESTED: Wrkforce: A proposal to help the Pkt Community

mikehostetler:
Please find our proposal for Wrkforce, software to help grow the Pkt Community in this PR.

### DIFF
--- a/projects/2021_12_14-wrkforce.md
+++ b/projects/2021_12_14-wrkforce.md
@@ -8,8 +8,9 @@
 - Projected duration: _3 months_
 - Projected effort: _4 people over 3 months_ - 1436 Labor Hours
 - Pre-project effort: _2 persons/4 months_
-- Requested PKT contribution: _11,000,000 PKT (@$0.02 -> $220,000)_
-  - 8,500,000 PKT for development of the project including all legal, accounting, and administrative needs.
+- Requested PKT contribution: _$195,000 USD_
+  - $195,000 USD for development of the project
+  - Legal, accounting, and administrative will be covered under this amount
   - 2,500,000 PKT for funding of community projects.
 - PKT address to pay to: _pkt1qtge3dqvuscp8v2x0xewyjj5ttutykjcvwq4l65_
 
@@ -132,7 +133,7 @@ Additional Pre-Project Effort:
 - Qualified financial auditor relationship established.
 - PKT Marketing Plan developed by Tom Varney to be used to prioritize & create work templates to execute on the plan.
 
-After the project is accepted, the Network Steward will pay 1,500,000 PKT
+After the project is accepted, the Network Steward will pay $30,000 USD
 
 ### Milestone 1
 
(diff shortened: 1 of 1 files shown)


cjdelisle:
Regarding your Request For Guidance:

After reading over the proposal, we like the idea but we have a few specific concerns which, if addressed, would in our opinion make a significantly stronger proposal.

## More agility
The structure of the project is such that it appears as though all 5 milestones must be complete before the first sub-projects can be processed. This is envisioned as 3 months but from our experiences of previous projects, this can be as much as a year or even more.

Furthermore, without iterative test cycles in front of real users, there is no evidence that when the platform is complete, it will address the needs of the community.

As an alternative, we propose that the team might setup a platform such as Discourse forum, and author a set of proceedures for managing a mostly manual register/KYC/proposal/vote processes through the forum. Then as the first few proposal cycles pass, the team may collect feedback and create a plan for improving the technology to make the process more automatic.

This would, in principle, allow for first sub-projects to start within 2 to 3 weeks of kickoff.

## Attention to prior art
There seems to be a significant number of other initiatives attempting to tackle this problem, in the proposal there was no mention of these projects. If the team has initimate knowledge of the strengths and weaknesses of these projects then a stronger proposal would be one which explicated why the team is proposing the platform that it is, rather than considering to adopt or modify one of the existing projects. If the team does not have such intimate knowledge, then a stronger proposal would be one which includes a research aspect in order to gain further insite to guide additional steps in the project.

Some examples of other vote funding projects include:
* https://docs.snapshot.org
* https://aragon.org/vocdoni
* https://www.boardroom.info
* https://gateway.metagov.org
* https://commonwealth.im
* https://www.loomio.com
* https://mirror.xyz/race
* https://otherinternet.substack.com/p/uniswap-interest-groups
* https://forum.cosmos.network/t/proposal-draft-use-the-forum-to-improve-governance-process/5530

## Clarify KYC and vote tallying processes
We observed that the vote tallying process is not exactly explained, in particular: Can one person vote on multiple projects in one round. If yes, is their weighted vote "split" between the projects they vote for or is it applied multiple times ([Approval voting](https://en.wikipedia.org/wiki/Approval_voting)). We consider approval voting to be simple and we do not identify any specific associated risks at this time, or likewise for any of the related multi-winner voting systems ([proportional approval voting](https://en.wikipedia.org/wiki/Proportional_approval_voting), [sequential proportional approval voting](https://en.wikipedia.org/wiki/Sequential_proportional_approval_voting)).

Secondly, the KYC process could be somewhat clarified to explain where peoples' personal data will be stored. In particular, if a 3rd party KYC provider is used, what information will that KYC provider be sharing with the team.

## Dollar-denomination and clear use of funds
At the time when the project proposal was written, the value was PKT denominated with a PKT price estimated at 0.02 USD. As of this writing wrapped PKT is trading at just under 0.04, so these budgets are no longer accurate.

Rather than trying to predict the value of PKT at the time when the work is done, we recommend specifying the costs in a national currency and then allowing that coins from will be sold in order to pay for the costs of the work as those costs are born. A rise in the price of the coin will allow for more funds to be allocated toward projects.

Furthermore, we think a stronger proposal would include at least a rough breakdown of projected costs, such as that spend on development, management, and legal, for each round of the project.

cjdelisle:
Unfortunately this project was not selected for funding. You are invited to look at the comments of the reviewers which can be found inside of the [Vote file](https://github.com/pkt-cash/ns-projects/blob/master/vote_2022_02_10.py). Wishing you good luck with proposing in the next round.

PR #89, opened 2022-02-14
Lets propose a new round

cjdelisle:
Lets propose a new round with 20mn budget and deadline 4th of March. WDYT ?


benhylau (approved):

cjdelisle:
Well, it looks like we have a quarum and I would like to have this land as quickly as possible to give people a chance, so I'm going to merge this now.

PR #93, opened 2022-02-28
Status update re sale of PKT

cjdelisle:
Providing a status update that PKT has been sold.


cjdelisle:
Merging this because it is a simple status declaration from the project.

PR #94, opened 2022-03-02
Adding proposal for Markdown Push Notifications and Mobile Apps

soricine:
Adding + Update proposal for Markdown Push Notifications and Mobile Apps

### DIFF
--- a/projects/2022_03_02_Mattermost_Push_Proxy_And_Moblie_Apps.md
+++ b/projects/2022_03_02_Mattermost_Push_Proxy_And_Moblie_Apps.md
@@ -1,20 +1,18 @@
-﻿
+
 # PKT Mattermost Push Proxy and Mobile Apps
 
--   Project Name: e.g. PKT Mattermost Push Proxy and Mobile Apps
+-   Project Name: PKT Mattermost Push Proxy and Mobile Apps
 -   Contact Email: soricine@gmail.com
 -   Project participants:
     -   Christofer Koutelieris (soricine@gmail.com) will work 50% of full time
     -   Adonis Gaitatzis (backupbrain@gmail.com) will work 50% of full time
-    _Other participants will be added as  needed_
-
--   Projected duration: e.g. 2 months
--   Projected effort: e.g. 2 person-months
--   Pre-project effort: e.g. 0.5 person/months
--   Requested PKT contribution: e.g. ADD ALL PKT PKT
+    -   _Other participants will be added as needed_
+-   Requested rate: 68 EUR / hour
+-   Total requested contribution: 15640 EUR
+-   Maximum requested PKT contribution: 600000 PKT
 -   PKT address to pay to: pkt1quxegcaslkacj75vu7kdhadm72dyu2kkt2axljn
+-   Repository with software which will be improved: e.g. https://github.com/soricine/pkt.chat-mattermost-utils
 - Project status: **PROPOSED**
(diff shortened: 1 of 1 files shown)


projects/2022_03_02_Mattermost_Push_Proxy_And_Moblie_Apps.md
```diff
@@ -0,0 +1,169 @@
...
+
+#### Push notification software running on a server
+
+At this step we will connect a web server, likely NGINX to the Mattermost push proxy software, so that the push proxy can operate on the public Internet over secure HTTPS.
+
+Estimated man-hours: 15
+
+#### Payout
+
+After a report is delivered on this milestone and it is approved by the NS, the NS will send 40,000 PKT.
```
cjdelisle, line 94: This is a personal comment and not the opinion of the NS.

I recommend that you specify the cost of each milestone in stable national currency, and report the trading price of PKT *when you deliver the milestone*. This way you will be more or less guaranteed to get paid fairly, and not over-charge in case of price instability.

You will still need establish a maximum cost in PKT of the entire project for NS accounting purposes, but this can be done by adding up your costs at a conservative PKT price.

You may consider looking at this template: https://cryptpad.fr/code/#/2/code/view/lHORne-EdOqRVL2kBQ7SCPpt8U9SoRF5utOCFjnaGVs/ for inspiration about how to structure a project in stable currency and then map it over to PKT.

cjdelisle:
I'm sorry, we are not able to consider updates after the deadline.

projects/2022_03_02_Mattermost_Push_Proxy_And_Moblie_Apps.md
soricine: Thankxx for the feedback, I just made the changes .

cjdelisle:
Congratulations, this project has been selected for funding.

As a point of order, changes to this proposal which were made *after the submission deadline* can not be considered, so the version of this project which has been accepted is found here: https://github.com/pkt-cash/ns-projects/blob/master/projects/2022_03_02_Mattermost_Push_Proxy_And_Moblie_Apps.md

Due to a vagueness regarding the cost of the project, we have assumed that the total cost is intended to be 600,000 PKT, but as this is not the sum of the milestones, I have edited the project proposal to reduce the final milestone payout from 266,666 to 253,334 in order to stay under the 600,000 limit.

If this is not what you want the please let us know in a comment and we can then remove the project and allow you to propose in the next round.

In order to tune your proposals in the future, it is recommended to consider the notes which were given by the reviewers: https://github.com/pkt-cash/ns-projects/blob/master/vote_2022_02_10.py

Please let me know as soon as possible whether you still agree to go forward with this project, and if so then wishing you a good project!

Caleb

soricine:
Hello there ,
First of all thank you very much for the funding for the : Markdown push
notification and Mobile Apps .
I still agree to go forward with this project .
Imnot sure about the procedure from here , so in any case you need
something extra from me to begin please send to me , else i will start the
project asap .

Best regards ,
Christofer Koutelieris

cjdelisle:
As there is no first kickoff payout, nothing else is required. Thanks.

https://github.com/pkt-cash/ns-projects/blob/master/projects/2022_03_02_Mattermost_Push_Proxy_And_Moblie_Apps.md#milestone-0-kickoff

PR #100, opened 2022-03-04
2022_03_03_AIoT_Punks_Atomic_Swap_Phase_1.md

AIoTPunks:


cjdelisle:
Congratulations, this project was selected for funding.

There was not an explicit mention of the maximum amount of PKT to be requested as is required in order to perform the evaluation, so the NS considered this project as costing 504000 PKT. I have updated the project to indicate this.

Please let me know as soon as possible if you agree with this, if not then you can withdraw the proposal and re-submit in the next round.

If you accept, then here's wishing you an excellent project!

Caleb

PR #108, opened 2022-03-15
Vote bugfix

Frank-Buss:
Currently the vote script doesn't work anymore, because of a bug in rrv_vote.py. This PR fixes the bug, and shows the individual votes for the score.

It worked with the previous votes, because the sum of all filtered projects was higher than the available budget. This time (with the current votes), the sum of the project budges of all filtered projects is less than the available budget.


cjdelisle:
In theory this represents a change of policy, but the original vote tally algorithm causes a crash if not all of the budget is spent, because there are projects deemed unacceptable.
This patch does not make any substantive change except to exit early with projects which have already been selected for acceptance, therefore I think it remains entirely in the spirit of the voting as it was originally defined. I will tentatively accept this now pending any objection from @Arceliar who originally wrote the algorithm.

PR #111, opened 2022-04-04
Project Kickoff Payment Request

vk-vinay:
Considering network steward votes and comments  project milestone budget restructuring made in USD

thankyou


cjdelisle:
You cannot change the amounts that are being requested, if you don't agree to the project as it was submitted and reviewed then you'll have to withdraw and resubmit in the next round. We need a pull request with just the bottom section requesting the sum of PKT, but keeping the amounts as they were defined.

vk-vinay:
will create new PR asap

PR #113, opened 2022-04-04
Edit readme.md number #8 was used for two steps

tomas347:
The readme has two times the number 8 refered as step.


cjdelisle:
Thank you !

cjdelisle:
Thank you !

PR #114, opened 2022-04-11
Housekeeping: Log the advance of payments to PktPay and pktchat upgrade

cjdelisle:
I advanced these payments to the applicants personally to save the NS the need to make as many payments. Please pay me back when it is convenient to do so.


cjdelisle:
Going to accept this myself because it is only documentative.

PR #115, opened 2022-04-12
Propose a call for end of APR 2022 

cjdelisle:
* 10mn PKT round proposed.
* Newly updated template using national currency as base.
* Removed "project_ideas.md" which are largely outdated now.

Let me know if you guys think this is a good idea.


cjdelisle:
Closing this because it did not gain consensus

PR #125, opened 2022-08-23
Re-add the project template which was removed by a PR

cjdelisle:


cjdelisle:
Going to self-merge this because it should be non-controversial.

PR #127, opened 2022-09-19
milestone 3 and 4 report for pkt_vpn_utility 

backupbrain:


cjdelisle (approved):
We have a pretty functional Android application with a wallet and VPN. It's available on the playstore and in github under https://github.com/anode-co/AnodeVPN-android

As a wallet it is able to open/manage multiple separate wallet.db files, has a nice way of displaying sync status, and has undergone quite a few revisions, including a whole redesign of the UI mid-project.

Of course it also has an integrated cjdns node and VPN client capability.

It has been delivered for both android (phone/tablet) and also android TV (which I happen to know required a fair bit of extra work):
* https://play.google.com/store/apps/details?id=co.anode.anodium.playstore
* https://play.google.com/store/apps/details?id=co.anode.anodium.androidtv

There is also a dockerized version of the VPN server developed by PKT Pal: https://github.com/pkteer/PKT-Server

From the report, we can see some features which were deferred, or were developed and then later switched off. I think this is acceptable because the results show an obvious good faith effort, even above-and-beyond the mandated scope, to develop something that would be usable by the public.

For me this is acceptable.

Arceliar (approved):
It appears that all major requirements were either met (possibly after being deferred, but that's life), or de-prioritized when they were found to be at least partly redundant with features already provided by the target platform (e.g. Android already tracks app resource use).

I think that a good faith effort has been made to meet the milestone deliverables, and that "morally" speaking this has been successful, so I'm happy to approve.

backupbrain:
I believe I shouldn't review this since I wrote this.

neilalexander (approved):
Having spent a lot of time swearing at Android's `VpnService` API in the past, to get to the point that the app is stable and usable is an accomplishment worthy of some merit alone, even before adding user interface sugar. It seems that the app is delivering on the milestones so happy to approve.

cjdelisle:
Merging this request as it has been properly approved.

PR #129, opened 2023-05-20
Update 2019_12_14_pkt-whitepaper-website.md

josh-berger:
Here is the final asset to be delivered for this project. Please note the updated wallet address as well

### DIFF
--- a/projects/2019_12_14_pkt-whitepaper-website.md
+++ b/projects/2019_12_14_pkt-whitepaper-website.md
@@ -138,7 +138,7 @@ Videographer/camera/sound: PKT5,000,000
 * Posting of white paper on the website
 
 ## Payments
-All payments shall be made to p9E3mhUtpinkvrn71AmoHNCoWYzMEmCiKq
+All payments shall be made to pkt1q09t0a07wc7gxgpynkqcy7utwzuacc0hj7w3k07
 
 ## Project Milestones
 The following are milestones for the progress of the project by which the network steward can evaluate the success of the project.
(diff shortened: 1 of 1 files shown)


cjdelisle (approved):
This milestone entails the delivery of a video which is specified as follows:
```
Presentation video: Working with Jesse and Josh, I will create a video presentation of the deck to be displayed on the website. This video will help users understand the project and excite them to dig in deeper.
```

My reactions of the video are:
1. There was definitely honest effort put into this, it went well above and beyond the specified scope as they hired a voice actor and used stock footage.
2. It wasn't as information dense as it could have been, but I don't think that's a valid critique of whether it was delivered to spec.
3. It mentions internet being "free" which I was at first confused about, but in fact what it is referencing is the "connect for free, pay for priority" principle which is elaborated in the PKT Network whitepaper.
4. Lastly, it's clear to me that the applicant took the spirit of the mission seriously. They could have made a specific video for the purpose of the milestone which is just them reading the deck which has already been approved. It would have easily passed inspection but would be unlikely to achieve the objective to "help users understand the project and excite them to dig in deeper".

For me this is a good.

And the detail of change of the recipient address, this PR has been made by the same github user who created the project proposal from the beginning and nobody has commented any objection to it so it seems fine. Of note, it is a change from the legacy style address that is no longer used to the modern style address, so it seems like an upgrade of the wallet.

Arceliar (approved):
Personally, I think the video leans a little bit too much on trying to "excite" the viewer, and not enough on helping them "understand" what pkt is / how it works. It would be more likely to drive me away from the project (it comes off like it's trying to sell me something, so as soon as I pull up the project page and see that crypto is involved, I would assume it's part of yet another crypto scam), but I realize that is highly subjective and I may be in the minority here.

The requirements in the proposal were pretty vague to begin with, so I do think that the video meets them (any complaints I have could have been avoided if we had insisted on more precise requirements, or that a script etc be delivered as part of an earlier milestone, so if that's anyone's failing then it's ours).

I don't think it's realistic to request any changes to the video at this point, and I see no reason to doubt that a good faith effort was made, I think it's just a difference in taste between myself and the authors. Since the stated objective from the milestone was met, that's good enough to get my approval on this one.

neilalexander (approved):
Satisfied that the deliverables were met and that the video, while light on technical details, should be overall positive in the marketing of the project.

cjdelisle:
Merging this request as it has been properly approved.

PR #132, opened 2024-02-01
Proposing a new Network Steward vote system

cjdelisle:
Please review this proposal.

### DIFF
--- a/projects/2024_02_01_new_network_steward.md
+++ b/projects/2024_02_01_new_network_steward.md
@@ -1,0 +1,126 @@
+# New Network Steward Vote
+
+* Project Name: New Network Steward Vote
+* Contact Email: cjd@cjdns.fr
+* Project participants:
+  * Caleb James DeLisle
+  * Whoever wins the vote
+* Projected duration: 10 weeks
+* Projected effort: I don't expect payment for this
+* Pre-project effort: About 2 person/months, but I don't expect payment for this
+* Requested PKT contribution: 16 million PKT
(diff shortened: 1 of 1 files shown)


cjdelisle:
I would like to provide a status update on this project:

1. I have developed a simple GUI based voting tool which allows people to cast votes using the private keys exported from their PKT wallets - so people with any kind of wallet will be able to vote. You can see this tool here: https://github.com/cjdelisle/PKT-Voter and you can see instructions for its use here: https://twitter.com/cjdelisle/status/1758198532870217927

2. I have created an updated block explorer backend, and the author of PacketScan has created a frontend for exploring the votes: https://packetscan.io/election and we can see currently there are 22 voters and 12 candidates with a total value of 8 million PKT.

3. The author of PacketScan has announced his candidacy here: https://pkt.watch/election/ to use the proceeds for building a web application for facilitating the Network Steward process for whomever wins in the future.

backupbrain (approved):
I was able to review the Electronium tests and the [pktd-fullnode code](https://github.com/pkt-cash/PKT-FullNode).

From what I read, the code makes sense and is in line with the proposed voting procedures.

As I understand the code, there is a voting period defined in a range of blocks, and votes are calculated through the counting of money deposited during the voting period into candidate addresses.

Arceliar (approved):
I have some minor concerns about some details / failure modes of the voting system, but I don't know that they're fixable in any system that involves delegation / proxy voting. There are good technical reasons not to use a more complex ballot, so this is probably a reasonable compromise.

So this looks like a good path forward from my side.

cjdelisle:
This has been approved.

PR #134, opened 2024-02-27
Updated the new network steward project with information about the new voting system

cjdelisle:


cjdelisle:
I'm going to self-merge this directly because it's just housekeeping and should not be controversial.

